PR #2, opened 2020-05-01
[G2M] WidgetStats to unify number and trend

woozyking:


src/widget-stats.js
```diff
@@ -0,0 +1,74 @@
...
+    padding: theme.spacing(2),
+    textAlign: 'center',
+    height: '100%'
+  },
+  title: {
+    padding: '8px'
+  }
+}))
+
+const WidgetStats = ({ title, value, prev, isPercentage, children }) => {
```
DoParkEQ, line 22: What do you think about changing `prev` to `trendVal`?

woozyking: `prev` is more generic, as `value` represent `current` value

in this component's actual usage, you can provide more suggestive variable naming (such as `trendVal`):

```jsx
<WidgetStats value={currentVal} prev={trendVal} />
```

DoParkEQ: Makes sense

PR #5, opened 2020-05-04
[G2M] Refactor - add thousand operator on values

tlowande:
>add thousand operator on: 
- WidgetTrend
- WidgetStats
- Table
> create storybook mockup data for table

### DIFF
--- a/src/widget-stats.js
+++ b/src/widget-stats.js
@@ -30,12 +30,12 @@ const WidgetStats = ({ title, value, prev, isPercentage, children }) => {
       )
     }
     if (prev != null) {
-      const trend = Math.round(((value - prev) / prev) * 100)
+      const trend = Math.round(((value - prev) / prev) * 100).toLocaleString()
       return (
         <Typography variant='body1'>
           {trend > 0
-            ? <TrendingUpRoundedIcon style={{ color: 'red' }}/>
-            : <TrendingDownRoundedIcon style={{ color: 'green' }} />
+            ? <TrendingUpRoundedIcon style={{ paddingRight: '8px', color: 'red' }}/>
+            : <TrendingDownRoundedIcon style={{ paddingRight: '8px',color: 'green' }} />
           }
           {trend}%
           {trend > 0 ? ' increase ' : ' decrease '}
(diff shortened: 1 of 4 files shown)


stories/data-table.stories.js
```diff
@@ -4,8 +4,22 @@ import { storiesOf } from '@storybook/react'
...
+  {new_cases: 2, total_cases: 11, province: "Yukon", rate: 22},
+  {new_cases: 1, total_cases: 118, province: "New Brunswick", rate: 1},
+  {new_cases: 1, total_cases: 26, province: "PEI", rate: 4},
+  {new_cases: 1, total_cases: 5, province: "Northwest Territories", rate: 25},
+]
 
 storiesOf('DataTable', module)
   .add('Default', () => (
-    <DataTable />
+    <DataTable data={data}/>
```
DoParkEQ, line 24: It causes lint error because there is no data coming in with this repo

tlowande: there's no error, I just created a variable with the mockup data. Can you screenshot the error you see?

DoParkEQ: ah okay, didn't see that you created a mock data for this PR.

src/widget-stats.js
```diff
@@ -30,12 +30,12 @@ const WidgetStats = ({ title, value, prev, isPercentage, children }) => {
       )
     }
     if (prev != null) {
-      const trend = Math.round(((value - prev) / prev) * 100)
+      const trend = Math.round(((value - prev) / prev) * 100).toLocaleString()
```
tlowande, line 33: does it matter if `.toLocaleString()` is here instead of on `line 40`?

stories/data-table.stories.js
woozyking: could you carry this into a different story and leave `Default` intentional for "defaultProps" case.

```js
storiesOf('DataTable', module)
  .add('Default', () => (
	<DataTable />
  ))
  .add('with data', () => (
	<DataTable data={data} />
  ))
```

src/widget-stats.js
woozyking: logically speaking, `trend` should be kept as-is (without `toLocaleString()`) since it's used for comparisons.

however, knowing how JS works in mysterious ways when it comes down to comparison, with the current `trend > 0` use case, it's probably _okay_.

so at this point it's up to you.

PR #6, opened 2020-05-05
[G2M] modified % mechanism

DoParkEQ:


stories/data-table.stories.js
```diff
@@ -21,5 +21,6 @@ const data = [
 
 storiesOf('DataTable', module)
   .add('Default', () => (
-    <DataTable data={data}/>
+    <DataTable data={data} 
+      isPercentage={true} whichColumn={3}/>
```
tlowande, line 25: _pref_: leo once asked me to keep the default with no data, so I updated that on the merged version

PR #15, opened 2020-05-15
[G2M] Added initial tests using snapshots

gulsahsarsilmaz:


__tests__/data-table/data-table.js
```diff
@@ -0,0 +1,20 @@
...
+describe('<DataTable/>', () => {
+  it('should render successfully', async () => {
+    let data = {
+      data: [],
+      isPercentage: false
+    }
+
+    const component = renderer.create(
+      <DataTable {...data} />
+    )
```
woozyking, line 15: we can use `test.each` for matrix-testing for props permutation:

```js
test.each([
  {}, // no props, all default
  { data: [] }, // only data
  { data: [], isPercentage: false }, // both data and isPercentage
  ...other combinations of props
])('%p', (props) => {
  const component = renderer.create(
    <DataTable {...props} />
  )
  // ...expect stuff
})

```

PR #18, opened 2020-05-19
[G2M] Added Additional unit tests to data table and widget for validity

gulsahsarsilmaz:
Added additional prop testing for Data Table and Widget Trend.

### DIFF
--- a/__tests__/data-table/data-table.js
+++ b/__tests__/data-table/data-table.js
@@ -3,6 +3,17 @@ import renderer from 'react-test-renderer'
 import DataTable from '../../src/data-table';
 
 
+beforeEach(() => {
+  jest.spyOn(console, 'error')
+  jest.spyOn(console, 'warn')
+})
+
+afterEach(() => {
+  /* eslint-disable no-console */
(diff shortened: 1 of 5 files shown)


__tests__/data-table/data-table.js
```diff
@@ -17,4 +17,30 @@ describe('<DataTable/>', () => {
     const tree = component.toJSON()
     expect(tree).toMatchSnapshot()
   })
+
+  // Failing Cases
```
woozyking, line 21: why are these considered failing cases?

gulsahsarsilmaz:
```
There are only two hard things in Computer Science: cache invalidation and naming things.
Phil Karlton
```
😆 I couldnt name them properly.

woozyking: perhaps "defaultProps" cases?

gulsahsarsilmaz: - [ ] Update comment on first Default Props - Missing Cases
- [ ] Update comment on second Default Props - Valid Cases

PR #20, opened 2020-05-20
[G2M] DynamicButton - change onClick propType to func

Goondrious:


src/dynamic-button.js
```diff
@@ -61,7 +61,7 @@ DynamicButton.propTypes = {
   disabled: PropTypes.bool,
   addIcon: PropTypes.node,
   onWhichSide: PropTypes.string,
-  onClick: PropTypes.node,
+  onClick: PropTypes.func,
```
woozyking, line 64: make sure to have a `defaultProps` for this one as well (and all other not `isRequired` propTypes)

```jsx
DynamicButton.defaultProps = {
  onClick: () => {}, // this would do
  // or
  onClick: null, // this works too
}
```

PR #25, opened 2020-05-29
[G2M] Fix hardcoded Widget Stats

maluhoss:
- made Widget Stats component more flexible
- include props for units, text and trend info like whether the trend should show as a percentage and whether increase is green or red

### DIFF
--- a/src/widget-stats.js
+++ b/src/widget-stats.js
@@ -19,7 +19,7 @@ const useStyles = makeStyles((theme) => ({
   }
 }))
 
-const WidgetStats = ({ title, value, prev, isPercentage, children }) => {
+const WidgetStats = ({ title, value, prev, isPercentage, children, comparedTo }) => {
   const classes = useStyles()
   const renderTrend = () => {
     if (children) {
@@ -39,7 +39,7 @@ const WidgetStats = ({ title, value, prev, isPercentage, children }) => {
           }
           {trend}%
           {trend > 0 ? ' increase ' : ' decrease '}
-          from before
+          {comparedTo}
         </Typography>
       )
     }
@@ -61,6 +61,7 @@ WidgetStats.propTypes = {
   prev: PropTypes.number,
   isPercentage: PropTypes.bool,
   children: PropTypes.object,
+  comparedTo: PropTypes.string,
 }
 
 WidgetStats.defaultProps = {
@@ -69,6 +70,7 @@ WidgetStats.defaultProps = {
   prev: null,
   isPercentage: false,
   children: null,
+  comparedTo: null
 }
 
 export default WidgetStats
--- a/src/widget-stats.js
+++ b/src/widget-stats.js
@@ -19,38 +19,30 @@ const useStyles = makeStyles((theme) => ({
   }
 }))
 
-const WidgetStats = ({ title, value, prev, isPercentage, children, comparedTo }) => {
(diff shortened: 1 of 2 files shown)


src/widget-stats.js
```diff
@@ -19,38 +19,30 @@ const useStyles = makeStyles((theme) => ({
...
-        </Typography>
-      )
-    }
-    return null
-  }
+  const { isTrendPercentage, upIsGreen, comparedTo, up, down } = trendInfo
+  const trendColours = (upIsGreen) ? ['green', 'red'] : ['red', 'green']
+  const trend = (isTrendPercentage) 
+    ? Math.round(((value - prev) / prev) * 100).toLocaleString()
+    : Math.round(value - prev)
```
tlowande, line 28: _comment_: keep `.toLocaleString()` even if percentage is false

src/widget-stats.js
```diff
@@ -19,38 +19,30 @@ const useStyles = makeStyles((theme) => ({
...
 
   return (
     <Paper className={classes.paper} variant='outlined'>
       <Typography className={classes.title} variant='subtitle2' gutterBottom>{title}</Typography>
-      <Typography variant='h5' gutterBottom>{value.toLocaleString()}{`${isPercentage ? `%` : '' }`}</Typography>
-      {renderTrend()}
+      <Typography variant='h5' gutterBottom>{value.toLocaleString()}{`${units}`}</Typography>
```
tlowande, line 33: _comment_: add storybook with `units` too

src/widget-stats.js
```diff
@@ -19,38 +19,30 @@ const useStyles = makeStyles((theme) => ({
...
+      {children && <Typography variant='body1'>
+        {children}
+      </Typography>}
+      {prev && <Typography variant='body1'>
+        {trend > 0
+          ? <TrendingUpRoundedIcon style={{ paddingRight: '8px', color: trendColours[0] }}/>
+          : <TrendingDownRoundedIcon style={{ paddingRight: '8px',color: trendColours[1] }} />
```
tlowande, line 40: _comment_: not sure about the use cases of this widget for atom/locus but it comes to mind now that the icon could also be optional and passed it over to the widget (although it makes props huge) and this could be the default icon.

Goondrious:
> _question_: Should there be new component tests though?

they're mostly flexible/superficial, but the last one will fail, I think: https://github.com/EQWorks/react-labs/blob/master/__tests__/widget-trend/widget-trend.js

src/widget-stats.js
maluhoss: This seems like it'll have to be done on another PR. Not as straightforward to dynamically load material-ui icons. Atleast that was my take on making icon flexible.

maluhoss:
> > _question_: Should there be new component tests though?
> 
> they're mostly flexible/superficial, but the last one will fail, I think: https://github.com/EQWorks/react-labs/blob/master/__tests__/widget-trend/widget-trend.js

I deleted this one so it should be good.

Goondrious:
Oh, shoot, that was my mistake in approving, then.

I'm not very experienced with tests, but as we start adding them more, we shouldn't be deleting old tests unless they're obsolete; we should extend them to match the new case.

These were "basic" from understanding: just "will it render" via snapshot and 2 pretty superficial prop tests, but still should be wary of deleting.

Since it's not a habit of ours across repos, let's try to keep on each other about adding tests and good test practices.

PR #31, opened 2020-06-08
[G2M] react-table + MUI Table

woozyking:


src/table/table-column.js
```diff
@@ -0,0 +1,5 @@
+import { Component } from 'react'
+
+class TableColumn extends Component { }
+
+export default TableColumn
```
tlowande, line 5: is this a placeholder?

woozyking: it's a dummy component to enable the declarative way of passing columns:

```jsx
<Table data={data}>
  <Table.Column Header='Rate' accessor='rate' />
  <Table.Column Header='Name' accessor='name' />
</Table>

// equiv
<Table
  data={data}
  columns={[
    { Header: 'Rate', accessor: 'rate' },
    { Header: 'Name', accessor: 'name' },
  ]}
/>
```

handy for use-cases such as if we want to introduce additional components within the `<Table>` context, but not interfering with the core data table behaviors.

the columns props way, on the other hand, is easier for dynamic manipulation for columns configuration, for instance

```jsx
const columns = [
  { Header: 'Name', accessor: 'name' },
]
if (isSupaMon()) { // supa-mon exclusive column
  columns.push({ Header: 'Rate', accessor: 'rate' })
}

return (
  <Table data={data} columns={columns} />
)
```

PR #32, opened 2020-06-09
[G2M] List and Tabs Component

hyx131:


stories/tab-panels.stories.js
```diff
@@ -0,0 +1,51 @@
...
+
+const customTab = theme => ({
+  root: {
+    textTransform: 'none',
+    minWidth: 72,
+    fontFamily: [
+      'BlinkMacSystemFont',
+      '"Segoe UI"',
+      'Roboto',
+      '"Helvetica Neue"',
```
tlowande, line 26: _question_ why the 2-word ones have to have double quotes?

src/list/list-item.js
```diff
@@ -0,0 +1,153 @@
...
+  )
+}
+
+DefaultListItem.propTypes = {
+  itemSecondaryAction: PropTypes.any,
+  onClick: PropTypes.func,
+  selected: PropTypes.bool,
+  button: PropTypes.bool,
+  focusSelect: PropTypes.bool,
+  avatar: PropTypes.bool,
```
tlowande, line 137: _fyi_: I'm confused with `avatar` being Boolean, so I checked chrome inspect and there are quite a few warnings in there. Check it too. If you don't see it lemme know, I can screenshot them, they're just long.

stories/tab-panels.stories.js
hyx131: think its because the fontFamily array will be processed by linking all elements into a single string separated by a white space, so any names that contains a white space would need to be quoted to keep that space as a part of the name

src/list/list-item.js
```diff
@@ -0,0 +1,153 @@
...
+            {avatar}
+          </Avatar>
+        </ListItemAvatar>
+        }
+        <ListItemText
+          primary={itemHeading(heading, progressBar)}
+          secondary={details}
+        />
+        {itemSecondaryAction && <ListItemSecondaryAction>{itemSecondaryAction}</ListItemSecondaryAction>}
+        <Grid container xs={2} >
```
tlowande, line 109: _comment_: one of the warnings I mentioned is that `container` can't be used with sizes like `xs` but you can use `container` and `item` together to use `xs` just gotta check if that is what you're looking for

src/list/list-item.js
hyx131: thanks! will def need to revise through all the props (many of which may not be needed at all), so design wise also would also like to have @DoParkEQ 's opinion to decide on which prop/feature to keep/revise etc

src/list/list-item.js
```diff
@@ -0,0 +1,153 @@
...
+            <LinearProgress value={progressBar} variant='determinate' className={classes.linearProgressBar}/>
+          </Grid>}
+      </Grid>
+    )
+  }
+
+  return (
+    <>
+      <ListItem
+        style={{backgroundColor: open && '#f5f5f5'}}
```
woozyking, line 89: we should minimize inline style to have a clear, single location to modify look and feels from the style's definition such as the one through `makeStyles` from above, then apply conditions on `className` prop.

if more complex uses and combinations are needed, consider adopting https://www.npmjs.com/package/clsx

src/list/list-item.js
```diff
@@ -0,0 +1,153 @@
...
+        <Grid container xs={2} >
+          <Grid item container justify='flex-end' xs={12}>
+            {timeStatus ? <div><FiberManualRecord className={classes.complete} />&nbsp;&nbsp;{timeStatus} ago</div> : null}
+          </Grid>
+          <Grid item container justify='flex-end' xs={12} >
+            {expand
+              ? open
+                ? <IconButton disableFocusRipple disableRipple onClick={showDetails}><ExpandLess /></IconButton>
+                : <IconButton disableFocusRipple disableRipple onClick={showDetails}><ExpandMore /></IconButton>
+              : null}
```
woozyking, line 118: consider extracting out multi-level conditionals as a render function before the component `return`

```jsx
const renderIconButton = () => {
  if (!expand) {
    return null
  }
  return (
    <IconButton disableFocusRipple disableRipple onClick={showDetails}>
      {open ? <ExpandLess /> : <ExpandMore />}
    </IconButton>
  )
}
```

then in `return`

```jsx
<Grid item container justify='flex-end' xs={12} >
  {renderIconButton()}
</Grid>
```

src/list/list-item.js
```diff
@@ -0,0 +1,153 @@
...
+  },
+  inProgress: {
+    color: 'red',
+    width: '12px',
+    height: '12px',
+  },
+  linearProgressBar: {
+    borderRadius: 50
+  },
+  small: {
```
DoParkEQ, line 39: Is this style property for Avatar size?

src/list/index.js
```diff
@@ -0,0 +1,79 @@
...
+  const classes = useStyles()
+  const [selected, setSelected] = useState(button && focusSelect ? 0 : false)
+
+  const inputStyles = {
+    width: width
+  }
+
+  // move selected with data to parent level
+  return (
+    <List className={classes.root} style={{ ...inputStyles }} disablePadding>
```
DoParkEQ, line 39: A quick question, could you tell me why it uses inline style?

src/list/list-item.js
hyx131: ya

src/list/index.js
hyx131: was a place holder for styles passed down via props (potentially custom styles), e.g. the `width` property

src/list/index.js
```diff
@@ -0,0 +1,98 @@
...
+      ))}
+    </MUIList>
+  )
+}
+
+List.propTypes = {
+  divider: PropTypes.bool,
+  border: PropTypes.bool,
+  spacing: PropTypes.number,
+  data: PropTypes.array,
```
tlowande, line 77: isn't `data`: `PropTypes.array.isRequired` or not really?

src/list/list-item.js
```diff
@@ -0,0 +1,309 @@
...
+    )
+  }
+
+  return (
+    <div className={clsx({ [classes[spacing]]: spacing })}>
+      <MUIListItem
+        onClick={onClick}
+        selected={selected}
+        button={button}
+        {...buttonProps}
```
tlowande, line 210: what is this doing?  I couldn't find ripple in `MUIListItem` but maybe it works still!

src/tab-panels.js
```diff
@@ -0,0 +1,55 @@
...
+  customTab
+}) => {
+  const [value , setValue] = useState(tabIndex)
+  const onTabChange = (_, newVal) => setValue(newVal)
+  let TabsComponent = Tabs
+  let TabComponent = Tab
+
+  if (customTab && customTabs) {
+    TabsComponent = withStyles(customTabs)(Tabs)
+    TabComponent = withStyles(customTab)((props) => <Tab disableRipple {...props} />)
```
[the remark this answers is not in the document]
hyx131: good call! also for Tabs @DoParkEQ is currently designing its features so not sure if we need to keep this custom styling option at all

stories/list.stories.js
```diff
@@ -0,0 +1,246 @@
...
+    heading: 'Heading・Four',
+    progressBar: 0,
+    details: 'Secondary summary four.',
+    timeStatus: '',
+    expand: false,
+    expansionDetails: '',
+  },
+]
+
+storiesOf('List', module)
```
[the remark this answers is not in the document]
hyx131: nice~ will check it out!

stories/list.stories.js
```diff
@@ -0,0 +1,246 @@
...
+        { expand: 'end', timeStatus: '30m', progressBar: 80, progress: 'incomplete' },
+        { expand: 'end', timeStatus: '4h', progress: 'complete' }
+      ]
+      const newItem = {
+        ...item,
+        progressBar: newFields[i].progressBar,
+        timeStatus: newFields[i].timeStatus,
+        progress: newFields[i].progress,
+        expand: newFields[i].expand,
+        expansionDetails: 'Example expansion details written here.',
```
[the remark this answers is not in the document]
hyx131: omg i always make the same mistake and tend to write them all out 🤣 , good catch!

src/list/list-item.js
hyx131: its for the `disableRipple` button prop, so for listItems not being used as a `button`, this prop won't be passed to `listItem` component (creating side effects if `button` bool isn't `true` for mui's `listItem`)

stories/list.stories.js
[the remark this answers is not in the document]
DoParkEQ: New syntax is already used in some of `react-labs` component. I recommend using the latest syntax for future components!

src/list/list-item.js
```diff
@@ -0,0 +1,309 @@
...
+      height: theme.spacing(5)
+    },
+    lg: {
+      width: theme.spacing(7),
+      height: theme.spacing(7),
+      marginRight: theme.spacing(1),
+    },
+    1: {
+      border: `1px solid ${theme.palette.shade.secondary[300]}`,
+      borderRadius: theme.spacing(0.4),
```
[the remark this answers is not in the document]
DoParkEQ: Also instead of manually creating each style variant, you can pass spacing prop into makeStyles()

```
spacing: number => {
      border: `1px solid ${theme.palette.shade.secondary[300]}`,
      borderRadius: theme.spacing(0.4),
      marginBottom: theme.spacing(number),
},
(...)
const classes = useStyles(spacing)
```

DoParkEQ:
Screen Shot 2020-07-02 at 9 17 23 PM

Can we change the background color of the selected list to `theme.palette.shade.secondary[50]`?
Also, the body text has a different typeface.

stories/list.stories.js
```diff
@@ -0,0 +1,237 @@
...
+        divider
+        data={modifiedData} 
+      />
+    )
+  })
+  .add('Avatar Size & BgColor & Chip', () => {
+    const getAvatarData = (variant, color) => {
+      const modifiedData = defaultData.map((item ,i) => {
+        const newFields = [
+          { heading: 'Default', chip: 'Marketplace', chipColor: '#6fcf97' },
```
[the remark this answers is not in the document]
hyx131: could you provide an example of how to get, say `#f2c94c` from `useTheme()`?

src/list/list-item.js
```diff
@@ -0,0 +1,309 @@
...
+        {!avatar && expand !=='end' && renderIconButton()}
+        {renderAvatar()}
+        <ListItemText
+          primary={itemHeading(heading, progressBar)}
+          secondary={details}
+        />
+        <Grid item container xs={2}>
+          <Grid item container xs={12} justify='flex-start' alignItems='flex-end' direction='column'>
+            {chip && <div>
+              <Chip
```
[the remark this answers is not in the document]
hyx131: do you have any designs for `<Chip` in mind?

hyx131:
> Screen Shot 2020-07-02 at 9 17 23 PM
> 
> Can we change the background color of the selected list to `theme.palette.shade.secondary[50]`?
> Also, the body text has a different typeface.

updated

stories/list.stories.js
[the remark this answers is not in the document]
DoParkEQ: we can use theme properties directly inside of function with `useTheme()` However, we'll need to register colors in the theme object first.
```
const myFunction = () => {
    const theme = useTheme();
    (...) 
    chipColor: `${theme.palette.customColor.yellow}`,
    (...)

File: stories/list.stories.js [hunk 16]
```diff
@@ -0,0 +1,200 @@
...
+      data={defaultData} 
+    />
+  ))
+  .add('Expansion & ProgressBar & TimeStatus', () => {
+    const modifiedData = defaultData.map((item ,i) => {
+      const newFields = [
+        { expand: true, progressBar: 20, timeStatus: '30m', progress: 'incomplete' },
+        { expand: 'start', timeStatus: '2h', progress: 'complete' },
+        { expand: 'end', timeStatus: '30m', progressBar: 80, progress: 'incomplete' },
+        { expand: 'end', timeStatus: '4h', progress: 'complete' }
```
tlowande, line 43: Suggestion for the future:
- are the `expand` & `timeStatus` ever be used together but differently like this example? If yes, we could align things like text always start on the same position and `timeStatus` always be aligned no matter if it's a 3 letter word (like 30m) or 2 letter (like 4h)

image

YizhiCatherineZhang: I agree with this suggestion

YizhiCatherineZhang:
Also, this branch has a conflict in `CHANGELOG.md`

stories/list.stories.js
hyx131: noted!

PR #33, opened 2020-06-10
[G2M] Table - reach sui parity

woozyking:
reaching feature parity with https://eqworks.github.io/extended-semantic-components/?path=/story/datatable--default (usage interface/props are not entirely the same, architecture and presentation layer obviously vastly different too)

- column visibility toggle (with per `Table.Column` prop `noToggle` to lock)
- global filter (as search)
- pagination
- download (with split-button control for download all/visible-only)

published as ~~[0.6.0-alpha](https://www.npmjs.com/package/@eqworks/react-labs/v/0.6.0-alpha)~~ [0.6.1-alpha](https://www.npmjs.com/package/@eqworks/react-labs/v/0.6.1-alpha)


src/table/search.js
```diff
@@ -1,9 +1,10 @@
-import React from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 
 import InputBase from '@material-ui/core/InputBase'
 import SearchIcon from '@material-ui/icons/Search'
 import { fade, makeStyles } from '@material-ui/core/styles'
+import { useAsyncDebounce } from 'react-table'
```
tlowande, line 7: ❤️

src/table/download.js
```diff
@@ -0,0 +1,141 @@
...
+
+const options = ['Save all columns', 'Save only visible columns']
+
+const Download = ({ data, columns, visibleColumns }) => {
+  const anchorRef = useRef(null)
+  const [open, setOpen] = useState(false)
+  const [selectedIndex, setSelectedIndex] = useState(0)
+
+  const allowOptions = visibleColumns.length < columns.length && visibleColumns.length > 0
+  const saveVisible = selectedIndex > 0 && visibleColumns.length !== columns.length
```
tlowande, line 55: when you download all columns, `csv` content is `undefined` but if you choose only visible columns, the data comes in properly. I believe it's something around the `_cols` state that is being passed as this check seems alright at first glance.

PR #34, opened 2020-06-12
[G2M] table - revise download, toggle, incorp design zpl.io/29Q4lyp

woozyking:
design context https://zpl.io/29Q4lyp
published as https://www.npmjs.com/package/@eqworks/react-labs/v/0.6.2-alpha

known/related issues:
- currently loaded theme/typography through `.storybook/preview.js` renders weird font/icon sizes (and fontface too?)
- unable to use `styled-checkbox` as it has theme related crash


geoerika:
I have one question Leo! Wouldn't be cleaner if we don't show the  columns that we cannot toggle in the Filter column list? We cannot select/deselect them, so why show them in the first place?

woozyking:
> I have one question Leo! Wouldn't be cleaner if we don't show the columns that we cannot toggle in the Filter column list? We cannot select/deselect them, so why show them in the first place?

My opinions:

1. the column toggles have the same order as the columns shown on the table - if we don't show the `noToggle` ones, users may be either confused as to why some columns are not there, or they need to mentally picture the "gap" columns. Overall it's simpler to show matching columns, but indicate as disabled to indicate what can/cannot be done.
2. the column toggles is a part of the "Column filter" feature, which means it'll incorporate more than just visibility toggle, but per-column value filtering (such as "column A value greater than X").

tlowande:
could also update readme with `index.js` exports eventually

woozyking:
> looks and feels great with the theme and the filter! I see a few warnings on the console though mostly `propType` stuff
> image

think i should address it now

src/table/table-toolbar/toggle.js
```diff
@@ -0,0 +1,104 @@
...
+        {({ TransitionProps, placement }) => (
+          <Grow
+            {...TransitionProps}
+            style={{
+              transformOrigin: placement === 'bottom' ? 'center top' : 'center bottom',
+            }}
+          >
+            <Paper>
+              <ClickAwayListener onClickAway={handleClose}>
+                <List>
```
DoParkEQ, line 59: Just a question, is there any specific reason why it uses `<ListItem>` component over `<FormControl>`?

woozyking: it's not a form. i would use `Menu` and `MenuItem` over `List` if they were intended as Menu like the Download options (which would look very similar to this part if you compare)

src/table/index.js
```diff
@@ -168,7 +168,7 @@ const Table = ({ columns, data, children, downloadable, tableProps, headerGroupP
       ) : (
         <Card>
           <CardContent>
-            <Typography variant='p'>
+            <Typography component='p'>
```
DoParkEQ, line 171: How about `variant='body1`?

woozyking: guess you're not a fan of the margin-bottom :trollface:

tlowande: ah for margin-bottom there's a param called `gutterBottom` or the actual word `paragraph`

PR #36, opened 2020-06-16
[G2M] table - Table.hiddenColumns prop or Table.Column.hidden prop for init hidden state

woozyking:
also mild code refactor to group custom hook logic into `useTableConfig` custom hook

### DIFF
--- a/src/table/index.js
+++ b/src/table/index.js
@@ -48,27 +48,55 @@ const getHeader = (s) => [
   s.slice(1).replace(/_/g, ' '),
 ].join('')
 
-const Table = ({ columns, data, children, downloadable, tableProps, headerGroupProps }) => {
-  const classes = useStyles()
-  const [autoCols, setAutoCols] = useState(columns || [])
-  const _data = useMemo(() => data, [data])
+const useTableConfig = ({ data, hiddenColumns, children, columns }) => {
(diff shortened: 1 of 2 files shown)


src/table/index.js
```diff
@@ -48,27 +48,55 @@ const getHeader = (s) => [
...
+
+      // initial column hidden states
+      const _hidden = _columns.filter((c) => c.hidden).map((c) => typeof c.accessor === 'function' ? c.id : c.accessor)
+      setHidden(_hidden.length ? _hidden : (hiddenColumns || []))
     }
   }, [columns, data, children])
+
+  // memoized columns and data for useTable hook
   const _cols = useMemo(() => autoCols, [autoCols])
+  const _data = useMemo(() => data, [data])
```
tlowande, line 78: _question_: I know `react-table` requires the memoized values to avoid expensive computation and multiple re-renders, but I don't understand the overall use of `useMemo` mostly combined with `useEffect` since useEffect already has the dependency array that only recalculates if those values change anyways...

woozyking: good question. i just chose this way to be on the safe-side as that's how `react-table` main hook expects its input to be (there may not be many differences unless the dataset is much much larger)

PR #38, opened 2020-06-17
[WIP] theme fix

DoParkEQ:


src/themes/typography.js
```diff
@@ -1,52 +1,39 @@
-//The base pixel size is 8px.
 const typography = {
-  fontFamily: "Noto Sans",
+  fontFamily: 'Poppins',
```
YizhiCatherineZhang, line 2: Do we decided to change our font family?

DoParkEQ: No it is not supposed to be Poppins, but I am changing it to Open Sans

SamSverko:
Is this PR still needed, @DoParkEQ and @YizhiCatherineZhang ? The Typography was addressed and updated in PR #80 .

DoParkEQ:
It is no longer necessary as this is already addressed in another PR

PR #41, opened 2020-06-19
[G2M] Table - improvements

woozyking:
- hide toolbar when no title and no data
- Edit columns with maxHeight and auto overflow
- add initial sortBy capability
- (BREAKING) remove title property

published as ~0.8.0-alpha~ 1.0.0-alpha 🎊

### DIFF
--- a/stories/table.stories.js
+++ b/stories/table.stories.js
@@ -145,10 +145,6 @@ export const disableFilters = () => (
   />
 )
 
-export const title = () => (
-  <Table data={provinces} title='Cases by province' />
-)
-
 export const initialSortBy = () => (
   <Table
     data={provinces}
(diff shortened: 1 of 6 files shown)


stories/table.stories.js
```diff
@@ -149,6 +149,13 @@ export const title = () => (
   <Table data={provinces} title='Cases by province' />
```
DoParkEQ, line 149: I think this can be separated from the table component itself. I imagine Title/subtitle to live outside of card/paper component like this.
Screen Shot 2020-06-22 at 10 11 47 AM

woozyking: `title` property was added in [`65369a5` (#37)](https://github.com/EQWorks/react-labs/pull/37/commits/65369a5b4a78da06587702bef440d26084c4b47e)

woozyking: this property is now removed with version 1.0.0-alpha

PR #43, opened 2020-06-23
[G2M] table - react-table compatible selection-filter

woozyking:
selection-filter

published as 1.1.0-alpha

### DIFF
--- a/src/table/filters/selection-filter.js
+++ b/src/table/filters/selection-filter.js
@@ -1,0 +1,80 @@
+import React, { useMemo } from 'react'
+import PropTypes from 'prop-types'
+
+import List from '@material-ui/core/List'
+import ListItem from '@material-ui/core/ListItem'
+import ListItemIcon from '@material-ui/core/ListItemIcon'
+import ListItemText from '@material-ui/core/ListItemText'
+import Switch from '@material-ui/core/Switch'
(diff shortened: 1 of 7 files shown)


src/table/table-filter-label.js
```diff
@@ -28,7 +29,6 @@ const TableFilterLabel = ({ column }) => {
   const [open, setOpen] = useState(false)
 
   const handleClose = (e) => {
-    console.log('lol')
```
tlowande, line 31: lol! we all missed it!

src/table/filters/selection-filter.js
```diff
@@ -0,0 +1,80 @@
...
+        )
+      })}
+    </List>
+  )
+}
+
+SelectionFilter.propTypes = {
+  column: PropTypes.object.isRequired,
+}
+SelectionFilter.filterFn = (rows, id, filterValue) => {
```
tlowande, line 75: what this syntax is doing `SelectionFilter.filterFn`? (seems very object-oriented) like, creating a function inside the component - cause components have the object structure?

woozyking:
> Aesthetically I prefer the checkboxes over the toggles though

I was referring to the guideline on when to use [checkbox vs. switch](https://uxplanet.org/checkbox-vs-toggle-switch-7fc6e83f10b8)

PR #45, opened 2020-06-23
[G2M] Login ui components

hyx131:
Four new components added:
`<Alert />`
`<Card />`
`<Loading />`
`<TextField />`


src/alert.js
```diff
@@ -0,0 +1,39 @@
...
+        <MUIAlert severity={severity} {...props}>{message}</MUIAlert>
+      )}
+    </div>
+  )
+}
+
+const propTypes = {
+  severity: PropTypes.string.isRequired,
+  message: PropTypes.string.isRequired,
+  header: PropTypes.string,
```
woozyking, line 35: would want to provide a `defaultProps` value for all that's not `isRequired`

src/loading.js
```diff
@@ -0,0 +1,207 @@
...
+      color: '#c8cbcf'
+    }
+  }
+})
+
+/* eslint-disable react/prop-types */
+const ProgressWithLabel = ({ action, labelStyle, ...props }) => {
+  const classes = useStyles()
+  const actionType = action && action.split(' ')[0]
+  if (actionType === 'circular') return (
```
tlowande, line 75: [cool thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) I learned from kevin:
> action.startsWith('circular')

so no need to split and get the first

src/loading.js
```diff
@@ -0,0 +1,207 @@
...
+    <Box position='relative' display='inline-flex'>
+      <CircularProgress variant='static' {...props} />
+      <Box top={0} left={0} bottom={0} right={0} position='absolute' display='flex' alignItems='center' justifyContent='center'>
+        <Typography style={labelStyle} variant='caption' component='div' color='inherit'>{`${Math.round(props.value)}%`}</Typography>
+      </Box>
+    </Box>
+  )
+  if (actionType === 'linear') return (
```
tlowande, line 83: so you could just do `else` here!

src/loading.js
```diff
@@ -0,0 +1,207 @@
...
+          </Grid>
+        </Backdrop>
+      )
+    }
+  }
+
+  /* Loading as loading wrapper with skeleton */
+  if (skeletonConfig) return (
+    <div>
+      {!open && cloneElement(children, { ...childProps })}
```
tlowande, line 163: oh didn't know about `cloneElement` 🥇

stories/card.stories.js
```diff
@@ -0,0 +1,76 @@
...
+  </>
+)
+
+storiesOf('Card', module)
+  .add('Default', () => (
+    <Card
+      actionSide='start'
+      cardTitle='Interesting Title'
+      cardContent='Some content goes here..'
+      cardAction={<DynamicButton />}
```
tlowande, line 42: should all of this be `required` on propTypes validation if this is default or make them `defaultProps`??

src/alert.js
```diff
@@ -0,0 +1,39 @@
...
+  width: PropTypes.string,
+  height: PropTypes.string,
+}
+
+Alert.defaultProps = {
+  header: '',
+  width: '100%',
+  height: '',
```
DoParkEQ, line 36: I would recommend `height: 'auto'`

src/loading.js
```diff
@@ -0,0 +1,207 @@
...
+        <LinearProgress {...props} />
+      </Box>
+      <Box minWidth={35}>
+        <Typography variant='body2' className={classes.linearProgressLabel}>{`${Math.round(props.value)}%`}</Typography>
+      </Box>
+    </Box>
+  )
+}
+
+const Loading = ({ open, backDrop, action, message, progress, childProps, children, skeletonConfig }) => {
```
DoParkEQ, line 95: I think we should give it a name with the more direct noun form such as `Loader` to match it with the rest of component names

stories/alert.stories.js
```diff
@@ -0,0 +1,46 @@
...
+      <div style={{ marginBottom: '10px' }}><Alert variant='filled' severity='warning' message='Warning message.' /></div>
+      <div style={{ marginBottom: '10px' }}><Alert variant='filled' severity='info' message='Info message.' /></div>
+      <div style={{ marginBottom: '10px' }}><Alert variant='filled' severity='success' message='Success message.' /></div>
+    </>
+  ))
+  .add('Success with header', () => (
+    <Alert severity='success' header='Msg Header' message='Success message.' />
+  ))
+  .add('Success no icon', () => (
+    <Alert icon={false} severity='success' message='Success message.' />
```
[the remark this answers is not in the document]
hyx131: no warning so far, which warning does it give?

src/card.js
```diff
@@ -89,7 +91,14 @@ const Card = ({
...
       </Grid>}
-    </MUICard>
+    </>
+  )
+  if (Object.entries(styledContainerProps).length > 0) return (
+    <StyleCardContainer {...styledContainerProps}>{cardBody}</StyleCardContainer>
+  )
+
+  return (
+    <MUICard style={dimensions} className={classes[size]}>{cardBody}</MUICard>
```
DoParkEQ, line 101: Let's keep default container as `<StyledCardContainer/>`

src/card.js
```diff
@@ -126,6 +136,7 @@ Card.defaultProps = {
   alignContent: 'start',
   variantTitle: 'h5',
   variantContent: 'body2',
+  styledContainerProps: {},
```
DoParkEQ, line 139: It can be `...StyledCardContainer.defaultProps`

src/loader.js
```diff
@@ -0,0 +1,203 @@
...
+  )
+  return (
+    <Box position='relative' display='inline-flex'>
+      <CircularProgress variant='static' {...props} />
+      <Box top={0} left={0} bottom={0} right={0} position='absolute' display='flex' alignItems='center' justifyContent='center'>
+        <Typography style={labelStyle} variant='caption' component='div' color='inherit'>{`${Math.round(props.value)}%`}</Typography>
+      </Box>
+    </Box>
+  )
+}
```
woozyking, line 92: it's better to have this resided in its own module, with proper propTypes definitions too.

if it's very specific/local to `loader`, perhaps factor `load.js` -> `loader/index.js` and have this as `loader/progress-with-label.js`

src/loader.js
```diff
@@ -0,0 +1,203 @@
...
+          const progressDiff = Math.random() * 10
+          const bufferDiff = Math.random() * 10
+          setBufferProgress(bufferProgress + progressDiff)
+          setBuffer(bufferProgress + progressDiff + bufferDiff)
+        }
+      }
+    })
+    useEffect(() => {
+      const timer = setInterval(() => progressRef.current(), 500)
+      return () => clearInterval(timer)
```
woozyking, line 116: 👍

src/loader.js
```diff
@@ -0,0 +1,203 @@
...
+      {action === 'linear determinate label'
+        && <ProgressWithLabel action={action} variant='determinate' value={progress} />}
+      {action === 'linear buffer label'
+        && <ProgressWithLabel action={action} variant='buffer' value={bufferProgress} valueBuffer={buffer} />}
+      <Typography className={classes.linearBackdropText}>{message}</Typography>
+    </>
+  )
+
+  /* Loading as backdrop */
+  if (backDrop) {
```
woozyking, line 137: technically, it's one word so no need to camel-case it. `backdrop` should be fine.

src/alert.js
```diff
@@ -0,0 +1,39 @@
...
+
+import MUIAlert from '@material-ui/lab/Alert'
+import AlertTitle from '@material-ui/lab/AlertTitle'
+
+
+const Alert = ({ severity, message, header, width, height, ...props }) => {
+  const dimensions = { width, height }
+
+  return (
+    <div style={dimensions}>
```
woozyking, line 12: i wonder if there's a better way than inline style for this :thonk:

src/loader.js
```diff
@@ -0,0 +1,203 @@
...
+          <Typography>{message}</Typography>
+        </Grid>
+      </Backdrop>
+    )
+  }
+
+  /* Loading as loading wrapper with skeleton */
+  if (skeletonConfig) return (
+    <div>
+      {!open && cloneElement(children, { ...childProps })}
```
woozyking, line 159: what's the advantage of doing `cloneElement(children, { ...childProps })` than, say, leave `childProps` to be accepted by `children` itself?

such as: 

```jsx
<Loader open={open} action='circular'>
  <DynamicButton onClick={() => setOpen(true)} disabled={open}>Click Me</DynamicButton>
</Loader>
```

instead of

```
<Loader open={open} action='circular' childProps={{ disabled: open }}>
  <DynamicButton onClick={() => setOpen(true)}>Click Me</DynamicButton>
</Loader>
```

hyx131:
> For the `wapper-button`, we could have a better present.
> image
> The black loader above the button feels hard to read, to improve this, there are two ways we can learn from MUI or Semantic UI React.
> 
> 1. Make the button grey like MUI. https://codesandbox.io/s/o57l2
> 2. Remove the test like Semantic https://codesandbox.io/s/mt3v1?module=/example.js

@DoParkEQ 's `div` idea works great, now the buttons look like this when loading is on:

Primary Buttons:
image

Secondary Buttons:
image

Tertiary Buttons:
image

src/loader.js
hyx131: ...no advantage in this case lol, think this is the result from a rabbit hole at some point 🤔

src/alert.js
hyx131: do you have any suggestions to avoid this inline?

PR #46, opened 2020-06-23
Table - range filter

woozyking:


src/table/filters/range-filter.js
```diff
@@ -0,0 +1,110 @@
...
+      shortValue = shortValue.toFixed(1)
+    }
+    newValue = `${shortValue}${suffixes[suffixNum]}`
+  }
+  return newValue;
+}
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    width: '40ch',
```
tlowande, line 34: why using `ch`?

src/table/filters/range-filter.js
```diff
@@ -0,0 +1,110 @@
...
+  const classes = useStyles()
+  const [min, max] = useMemo(() => {
+    let min = preFilteredRows.length ? preFilteredRows[0].values[id] : 0
+    let max = preFilteredRows.length ? preFilteredRows[0].values[id] : 0
+    preFilteredRows.forEach(row => {
+      min = Math.min(row.values[id], min)
+      max = Math.max(row.values[id], max)
+    })
+    return [min, max]
+  }, [id, preFilteredRows])
```
tlowande, line 54: _comment_: I wonder if a reducer wouldn't be easier to read here

src/table/filters/range-filter.js
woozyking: because [`ch` reflects the font in-use](https://drafts.csswg.org/css-values-3/#ch). so it's a nice choice for input field related dimension definition

src/table/filters/range-filter.js
woozyking: this was a copy/paste from `react-table` example.

in my understanding, conceptually/functionally this would be the same as using `useState/useReducer` to manage the `min, max` pair, just that it would be more verbose. `useMemo`'s usage is justified in a way that it's a more direct path to define "computed state" based only on props, instead of using `useState/useReducer` + `useEffect` to achieve the same.

tlowande: I can see that, makes sense. But I was thinking about `preFilteredRows.reduce`

woozyking: oh i see what you mean

PR #49, opened 2020-06-29
[G2M] Table - remember me

woozyking:
rolling change on top of #48


src/table/index.js
```diff
@@ -90,10 +90,22 @@ const Table = ({
...
+  // remember me
+  const [
+    cachedHidden,
+    setCachedHidden,
+    removeCachedHidden,
+  ] = cached({ ...remember, key: `${remember.key}_HIDDEN` })(useState)(hidden)
+  const [
+    cachedSortBy,
+    setCachedSortBy,
+    removeCachedSortBy,
```
tlowande, line 107: _fancy, own repo, cool!_ This syntax of having `removeState`  is it from your library? Why not just `setCached(null)` or similar instead of removing it?

src/table/index.js
```diff
@@ -90,10 +90,22 @@ const Table = ({
...
+  ] = cached({ ...remember, key: `${remember.key}_SORT_BY` })(useState)(sortBy)
```
tlowande, line 108: _question_ why adding the option of having an expiration value? Don't you want the columns to always persist the way  user has chosen them?

stories/table.stories.js
```diff
@@ -191,3 +192,48 @@ export const tableProps = () => (
...
+      />
+    </>
+  )
+}
+
+export const rememberSortBy = () => {
+  const remember = {
+    key: 'DEMO_REMEMBER_SORT_BY',
+    ttl: 5, // remember for 5 minutes
+    sortBy: true,
```
tlowande, line 225: you can have both, `sortBy` and `hidden` under `remember`, right?

src/table/index.js
woozyking: https://woozyking.github.io/use-cached/?path=/story/cached--programmtic-cache-removal

it removes the cache (from localStorage) but keeps in-app/memory state intact

now that i think about it, this is probably better to be removed only when we unmount the component

src/table/index.js
woozyking: by default it's indefinitely cached. but it's nice to relay this option (for free) from the underlying lib.

stories/table.stories.js
woozyking: yes, you can have both.

src/table/index.js
YizhiCatherineZhang: Question @tlowande how do you know here has an expiration value? I cannot find where is the expiration in the code.

woozyking: it's documented here https://github.com/woozyking/use-cached/#cached-usestate

src/table/index.js
```diff
@@ -237,6 +278,9 @@ Table.defaultProps = {
   tableProps: {},
   headerGroupProps: {},
   sortBy: {},
+  remember: {
+    key: '__not_cached',
+  },
```
tlowande, line 283: default here @YizhiCatherineZhang

PR #51, opened 2020-07-02
[G2M] Styled Card Patterns

DoParkEQ:


src/styled-checkbox.js
```diff
@@ -95,7 +95,7 @@ StyledCheckbox.propTypes = {
 StyledCheckbox.defaultProps = {
   checked: false,
   onChange: null,
-  disabled: true,
+  disabled: false,
```
woozyking, line 98: all 3 of these props can be relayed directly from `...rest` as they're all native MUI `Checkbox` props.

stories/styled-card-container.stories.js
```diff
@@ -0,0 +1,235 @@
...
+  },
+  sub: {
+    color: theme.palette.grey[600],
+  },
+}));
+
+export const Default = () => {
+  const classes = useStyles();
+
+  const contents = ({ title, main, sub }) => {
```
woozyking, line 37: we can disable props validation for stories altogether, so safe to add that disable clause for this file

stories/data/cardInfo.js
```diff
@@ -0,0 +1,131 @@
...
+import img4 from "../../src/assets/fans.png";
+import img5 from "../../src/assets/rogers.jpg";
+import img6 from "../../src/assets/telus.jpg";
+import img7 from "../../src/assets/census.jpeg";
+import img8 from "../../src/assets/custom.jpg";
+import img9 from "../../src/assets/finance.jpg";
+import img10 from "../../src/assets/household.jpg";
+import img11 from "../../src/assets/persona.jpg";
+import img12 from "../../src/assets/retail.jpg";
+import img13 from "../../src/assets/telco.jpg";
```
woozyking, line 13: are these assets only used for stories? if so, we should move them into `stories/assets` instead to not inflate the library package size upon bundling/publishing

src/styled-checkbox.js
DoParkEQ: Does that mean that those can be removed?

woozyking: yea both `propTypes` and `defaultProps` can be removed.

you could keep the `defaultProps` if you wish to alter their default behaviour if they're different from MUI

src/styled-card-container.js
```diff
@@ -0,0 +1,93 @@
...
+      "&:hover": {
+        transition: "all .6s",
+        boxShadow: theme.palette.shadow[30],
+        backgroundSize: "105% auto",
+      },
+    },
+  };
+});
+
+const StyledCardContainer = ({ pattern, clickable, onClick, children }) => {
```
hyx131, line 62: think just `onClick` here would be sufficient

DoParkEQ: What about the default prop settings for onClick? Without clickable, it will always fire the default alert msg.

hyx131: empty function would be okay, like this `() => {}`, correct me if this isn't good practice :)

DoParkEQ: maybe @woozyking can give us some insights on this.

woozyking: depending on the use cases, having `defaultProps` for function `propTypes` can either be no-op function like @hyx131 said, or `null`.

if you need to invoke the function explicitly, a no-op function is needed:
```jsx
<Component
  // this `onClick` needs to be a function, so defaultProps should be () => {}
  onClick={(event) => { onClick(event) }}
/>
```

if you feed that into a native/react virtual event handler, default `null` is safe:
```jsx
<Component
  // onClick event handler accepts null safely
  onClick={onClick}
/>
```

stories/data/cardInfo.js
```diff
@@ -0,0 +1,131 @@
+import img1 from "../assets/branding.jpg";
```
woozyking, line 1: please keep the files all lower case and use `-` to join words, as suggested https://github.com/EQWorks/common/blob/master/naming.md#code-and-files-associative-naming (reasons within too)

DoParkEQ: such as `img-1`?

woozyking: sorry i meant this file `cardInfo.js` should be `card-info.js`

src/styled-card-container.js
DoParkEQ: So, is it correct that propTypes would be `number` and default props would be `null`?

woozyking: > So, is it correct that propTypes would be `number` and default props would be `null`?

also depends on the use cases, suppose prop name `value` has a type of `number` that's not required.

then in this case when you use it

```jsx
const Component = ({ value }) => {
  const renderFancyValue = () => {
    // if the prop is intended for arithmetics, then it's better to give it a numeral default value (or make it required)
    const calculated = 128 / value

    return (<FancyValue>{caluclated}</FancyValue>)
  }
  return (
    <Paper>
      // if value is used for truthy determination, then defaultProps value to null is better
      {value && renderFancyValue()}
    </Paper>
  )
}
```

stories/styled-card-container.stories.js
```diff
@@ -0,0 +1,236 @@
...
+            </CardContent>
+          </StyledCardContainer>
+        </Grid>
+      ))}
+    </Grid>
+  );
+};
+
+const useStyles2 = makeStyles((theme) => {
+  console.log(theme);
```
YizhiCatherineZhang, line 68: Please remove the log if it's not necessary.

PR #53, opened 2020-07-03
[G2M] Carousel

DoParkEQ:


src/imageSlider/carousel.js
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <div>
+      <Slider {...settings} ref={getRef}>
+        {content}
+      </Slider>
+    </div>
+  );
+};
+
+Carousel.PropTypes = {
```
woozyking, line 30: the reason why linter complains about no propTypes definition is that this needs to be `Carousel.propTypes`

DoParkEQ: 🤦

src/imageSlider/index.js
```diff
@@ -0,0 +1,36 @@
...
+  );
+};
+
+ImageSlider.propTypes = {
+  topSectionContent: PropTypes.object,
+  carouselContent: PropTypes.object,
+}
+
+ImageSlider.defaultProps = {
+  topSectionContent: {
```
DoParkEQ, line 25: Both `topSectionContent` and `carouselContent` are being passed to `<Carousel/>` via `<ImageSlider/>`. In this case, do those props still need to have default props both at parent and child level?

DoParkEQ: @woozyking

woozyking: no. in this case you can take a look at https://github.com/EQWorks/react-labs/blob/master/src/table/table-toolbar/index.js#L69 and see how the parent component `<TableToolbar />` stacks in the prop-types definitions of its children components `<Download />` and `<Toggle />`.

so usually it's better to define the prop-types at children's levels and reuse them to compose into parent levels, when necessary.

src/css/carousel.css
```diff
@@ -0,0 +1,5 @@
+.slick-slide {
+	padding-right: 16px;
+	box-sizing: border-box;
+ }
+
```
woozyking, line 5: let's keep using css-in-js

DoParkEQ: I wasn't sure how to append style attribute to '.slick-slide' through JSS as `<Slider/>` as it is using an external library. Would you be able to show me how?

PR #54, opened 2020-07-03
[G2M] Table - fix init hidden, change pagination display logic

woozyking:


src/table/index.js
```diff
@@ -212,26 +212,30 @@ const Table = ({
...
-              inputProps: { 'aria-label': 'rows per page' },
-              native: true,
-            }}
-            onChangePage={(_, page) => { gotoPage(page) }}
-            onChangeRowsPerPage={({ target: { value }}) => {
-              setPageSize(Number(value))
-            }}
-          />
+          {/* TODO: this seems to be simplifiable */}
+          {(0 < rows.length && rows.length < data.length ? rows.length > pageSize : rows.length > 0) && (
```
hyx131, line 216: did a quick boolean check out of curiosity in the form 0 < r < d lol, seems to work, but not sure if it will behave differently in real project use cases 🤔

woozyking: try this in browser or node console

```js
>> const a = 5
>> 1 < a < 4 // what should this be?
```

PR #55, opened 2020-07-08
[G2M]Table - adjustments to overlord usage

tlowande:
* hide sort by icon when `disableSortBy: true`
* fix console warning: 
> 1. makes pagination part of table so console stops complaining that `td` can't be a child of `div`
> 2. gives `globalFilter` initial value of empty string so propTypes won't complain of `undefined`

- range filter: overlord usecases - float numbers and percentages

> 1. `percentage: bool` was added to `column` so range filter can adjust to format. Works for `inViewRate` and `WinRate` if passed in overlord children table
```
        getAriaValueText={percentage ? (value) => (value*100) : abbreviateNumber}
        valueLabelFormat={percentage ? (value) => (value*100) : abbreviateNumber}
        step={max - min <= 1 ? 0.1 : 1}
```
> 2. remove textfield cause it doens't work properly with percentage usecase. Available at #59


src/table/index.js
```diff
@@ -171,77 +172,83 @@ const Table = ({
...
-        </>
+                )
+              })}
+            </TableBody>
+
+            {/* TODO: this seems to be simplifiable */}
+            {(0 < rows.length && rows.length < data.length ? rows.length > pageSize : rows.length > 0) && (
+              <TableFooter>
+                <TableRow>
+                  <TablePagination
```
woozyking, line 219: oh so that's how you're supposed to use table pagination properly, somehow missed from the official MUI example

src/table/index.js
```diff
@@ -171,77 +172,83 @@ const Table = ({
...
-                    {headerGroup.headers.map((column, i) => (
-                      <TableCell
-                        key={i}
-                        className={classes.head}
-                        {...column.getHeaderProps(column.getSortByToggleProps())}
-                      >
-                        {column.render('Header')}
-                        <TableSortLabel {...column} />
-                        {column.canFilter && (<TableFilterLabel column={column} />)}
+        // <>
```
woozyking, line 180: we can delete this if not needed anymore

src/table/index.js
```diff
@@ -171,77 +172,83 @@ const Table = ({
           downloadable={downloadable}
           data={data}
           preGlobalFilteredRows={preGlobalFilteredRows}
-          globalFilter={globalFilter}
+          globalFilter={globalFilter || ''}
```
woozyking, line 175: this addresses the prop warning on `globalFilter` i suppose?

src/table/index.js
```diff
@@ -171,77 +172,83 @@ const Table = ({
...
+              {headerGroups.map((headerGroup, i) => (
+                <TableRow key={i} {...headerGroup.getHeaderGroupProps(headerGroupProps)}>
+                  {headerGroup.headers.map((column, i) => (
+                    <TableCell
+                      key={i}
+                      className={classes.head}
+                      {...column.getHeaderProps(column.getSortByToggleProps())}
+                    >
+                      {column.render('Header')}
+                      {column.canSort && (<TableSortLabel {...column} />)}
```
woozyking, line 193: :thonk: i see. somehow i believed it was there and that's why i wondered why sort icon was there even after `disableSortBy` presents on column options.

src/table/index.js
tlowande: they have examples where it is outside the `<MUITable>` I guess this is actually a browser check that complains

src/table/index.js
tlowande: it happens, good thing it is a simple solution :)

src/table/index.js
woozyking: so we can blame the official doc?

src/table/filters/range-filter.js
```diff
@@ -25,23 +24,26 @@ function abbreviateNumber(value) {
...
     width: '40ch',
-    padding: theme.spacing(1),
+    padding: theme.spacing(4, 2, 0, 2),
+    textAlign: 'center'
   },
   text: {
     marginLeft: theme.spacing(1),
     marginRight: theme.spacing(1),
-    width: '18ch',
+    width: '16ch',
```
woozyking, line 42: oh this class is no longer needed

src/table/index.js
tlowande: lol I guess

src/table/filters/range-filter.js
tlowande: oh you're right, not for this anyways

src/table/index.js
```diff
@@ -243,16 +242,15 @@ const Table = ({
...
+            <CardContent>
+              <Typography variant='body1'>
                 No visible columns
             </Typography>
-          </CardContent>
-        </Card>
-      )}
+            </CardContent>
+          </Card>
+        )}
```
woozyking: why the extra indentation :thonk:

tlowande: not sure I just lint it though

PR #56, opened 2020-07-09
[G2M] Export new components

YizhiCatherineZhang:
- [x] ImageSlider
- [x] StyleCardContainer
- [x] StyledCheckbox
- [x] StyledRadio
- [x] StyledSwitch


src/index.js
```diff
@@ -6,3 +6,8 @@ export { default as QuickFilters } from './quick-filters'
 export { default as DynamicButton } from './dynamic-button'
 export { default as DynamicDataTable } from './dynamic-data-table' // deprecated by Table
 export { default as Table } from './table'
+export { default as ImageSlider } from './image-slider'
+export { default as StyleCardContainer } from './styled-card-container'
+export { default as StyledCheckbox } from './styled-checkbox'
+export { default as StyledRadio } from './styled-radio'
+export { default as StyledSwitch } from './styled-switch'
```
DoParkEQ, line 13: could you add export { default as Carousel } from './image-slider/carousel' ?

src/index.js
```diff
@@ -6,3 +6,9 @@ export { default as QuickFilters } from './quick-filters'
...
```
woozyking, line 13: at some point we should drop the prefix like `Dynamic` and `Styled` and just go with `Button`, `CardContainer`, `Checkbox`.

DoParkEQ: I think we can omit prefix and whenever we encounter situation we mix MUI and our own, we can add prefix to mui components?

woozyking: > I think we can omit prefix and whenever we encounter situation we mix MUI and our own, we can add prefix to mui components?

preciselaay 🧐

YizhiCatherineZhang: Do we need to include the name change in this PR? If not let's merge this one and then @DoParkEQ can keep working on the marketplace main page.

woozyking: no need to include the name change. i'll make a backward compatible way of introducing it later

src/index.js
```diff
@@ -6,3 +6,9 @@ export { default as QuickFilters } from './quick-filters'
 export { default as DynamicButton } from './dynamic-button'
 export { default as DynamicDataTable } from './dynamic-data-table' // deprecated by Table
 export { default as Table } from './table'
+export { default as ImageSlider } from './image-slider'
+export { default as StyleCardContainer } from './styled-card-container'
```
YizhiCatherineZhang, line 10: @woozyking @DoParkEQ I just found I had a typo here, should be `StyledCardContainer`. Totally my bad...

PR #62, opened 2020-07-15
[G2M]Table - custom range filter

tlowande:
@DoParkEQ this would be the way to pass custom filters!

this is using dates so I added mockup data.

basically just create your UI and create your filter and pass it to the table.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "eslint-plugin-react": "^7.14.3",
     "eslint-watch": "^6.0.1",
     "jest": "^26.0.1",
+    "moment": "^2.27.0",
     "react": "^16.13.0",
     "react-dom": "^16.13.0",
     "react-test-renderer": "^16.13.1"
--- a/src/card.js
+++ b/src/card.js
@@ -10,7 +10,7 @@ import Grid from '@material-ui/core/Grid'
 import Typography from '@material-ui/core/Typography'
 
 import { palette, typography } from './themes'
-import { StyleCardContainer } from '.'
+import { StyledCardContainer } from '.'
 
 
 const useStyles = makeStyles((t) => {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "react-dom": "^16.13.0"
   },
   "dependencies": {
+    "moment": "^2.27.0",
```
woozyking, line 50: oof, that's a whopping 4.12MB in size we just added to this package. to put it into perspective, react-labs as of 1.5.0 is 133kB :trollface: 

can we try to work without `moment`, for the practical reason of not bloating this package up?

for this particular case, seeing it is only used for stories, let's move this into `devDependencies` so it doesn't get packaged.

stories/data/custom-range-filter.js
```diff
@@ -0,0 +1,77 @@
...
+    flexWrap: 'wrap',
+  },
+  textField: {
+    marginLeft: theme.spacing(1),
+    marginRight: theme.spacing(1),
+    width: 200,
+  },
+}));
+
+export function DateRangePicker({ column: { filterValue, preFilteredRows, setFilter, id } }) {
```
woozyking, line 18: while shopping for a date range picker, i came across https://material-ui-pickers.mui-org.now.sh/ (unreleased, alpha version of https://material-ui-pickers.dev/) which has an unstable [date range picker](https://material-ui-pickers.mui-org.now.sh/demo/daterangepicker) implementation

stories/data/custom-range-filter.js
```diff
@@ -0,0 +1,83 @@
...
```
DoParkEQ, line 24: I think it'd be better to match the component name and its file name. What do you think?

Issue #72, opened 2020-07-18
Add link passing mechanism to ImageSlider


DoParkEQ:
```
button: {
   text: (...),
   linkTo: (...),
}
```

PR #73, opened 2020-07-20
[G2M] ImageSlider linkTo prop added

DoParkEQ:
fixed https://github.com/EQWorks/react-labs/issues/72

### DIFF
--- a/src/image-slider/top-section.js
+++ b/src/image-slider/top-section.js
@@ -39,8 +39,8 @@ const TopSection = ({topSectionContent, customSlider}) => {
         <Typography variant="body1" gutterBottom>{paragraph}</Typography>
       </Grid>
       <Grid item className={classes.link}>
-        <Link href="#" variant="body1" color="primary">
-          {button}
+        <Link href={button.linkTo} variant="body1" color="primary">
+          {button.text}
         </Link>
         <IconButton
           classes={{ root: classes.sliderControl }}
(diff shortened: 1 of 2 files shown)


SamSverko:
Link for the top section looks good to me 👍 

There's no link needed for each image/card item?

src/image-slider/top-section.js
```diff
@@ -39,8 +39,8 @@ const TopSection = ({topSectionContent, customSlider}) => {
         <Typography variant="body1" gutterBottom>{paragraph}</Typography>
       </Grid>
       <Grid item className={classes.link}>
-        <Link href="#" variant="body1" color="primary">
-          {button}
+        <Link href={button.linkTo} variant="body1" color="primary">
+          {button.text}
```
woozyking, line 43: this would pose a breaking change (that means a major version bump).

or, we can practice a bit of backward compatibility here:

```jsx
<Link href={typeof button === 'object' ? button.linkTo : '#'} ...>
  {typeof button === 'object' ? button.text : button}
```

this way the existing API (shape of props) would not break/crash the usage of `<TopSection />`, and a major version bump can be saved for the future.

make sure to mark clearly for deprecation in the changelog tho

DoParkEQ:
@SamSverko If someone wants to attach a link to a card, they can append it directly onto its card content

src/image-slider/top-section.js
DoParkEQ: Is this the correct backward compatibility workflow?
revise prop passing mechanism and release a new version of `react-labs` --> update the version of `react-labs` in snoke --> update the prop fed to ImageSlider --> deprecate old way of prop passing mechanism in ImageSlider in `react-labs`

PR #74, opened 2020-07-21
[G2M] Fix #68 - Build Chip Component

### LINKED ISSUE #68: Chip component
Create `<Chip/>` component based on the given design below.
Screen Shot 2020-07-16 at 8 10 32 PM

The component comes with three variations in terms of functionality: static, toggleable, and clickable. You can find the original design with detailed specifications on Zeplin.

**Static**
A static chip works as a tag for items in `<Card/>`, `<List/>` and `<Table/>`. It should allow users to customize its color, as well as dynamically selects text color (between white and black) that has a higher contrast ratio to it. 
* Its default color should be a primary color defined in `theme`.
* It should come with both round and rectangular shapes.

**Toggleable**
The component should also work as a toggle button which can be paired with tables or other selection controls to filter data. Both unselected and selected modes have a hover state. 
* You will be able to find predefined style keys in `theme`.  `palette.state.hoverWhite` when the component is unselected and `palette.state.hoverColored` when it is selected. 
* Please look up `<DynamicButton/>` as a reference.

**Deletable**
Deletable chips are ones that disappear when clicked. They also have a hover state and it should have the same style keys as toggleable.

Few more details for all variations:
- The ripple effect should be disabled.
- We can keep MUI's default size variation.

SamSverko:


SamSverko:
Still working on this. It breaks when you change the Theme to `DefaultTheme`. Will probably need a walkthrough of that with you @DoParkEQ :)

src/chip.js
```diff
@@ -0,0 +1,125 @@
...
+      border: `1px solid ${theme.palette.primary.main}`,
+      color: theme.palette.primary.main,
+    },
+    '&:focus': {
+      backgroundColor: '#FFFFFF',
+      border: `1px solid ${theme.palette.shade.secondary[400]}`,
+      color: theme.palette.shade.secondary[900],
+    }
+  },
+  toggle: {
```
DoParkEQ, line 29: Toggleable chips would need styling for disabled state.
``` 
"&:disabled": {
        color: theme.palette.common.white,
        opacity: 0.5,
      },
```

src/chip.js
```diff
@@ -0,0 +1,125 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { makeStyles } from '@material-ui/core/styles'
+import MUIChip from '@material-ui/core/Chip'
+import clsx from 'clsx'
+
+import { palette } from './themes'
+
+const useStyles = makeStyles((theme) => ({
+  chip: {
```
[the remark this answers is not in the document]
SamSverko: Does that mean we need to use this code for every single component?

stories/chip.stories.js
```diff
@@ -0,0 +1,47 @@
...
+
+import { Chip } from '../src/index'
+
+export default {
+  title: 'Lab/Chip',
+  component: Chip,
+  args: {
+    label: 'chip',
+  },
+  argTypes: {
```
DoParkEQ, line 11: I noticed that some props separately exist in `chip.js`. Do you mind if we have all props under `argTypes`? I think having them in one place would be easier for us to review all the props.

src/chip.js
```diff
@@ -0,0 +1,95 @@
...
+  */
+  onDelete: PropTypes.func,
+  /**
+    * If true, the element shape will be rectangular.
+  */
+  rectangle: PropTypes.bool,
+  /**
+    * The variant to use.
+  */
+  variant: PropTypes.oneOf(['default', 'outlined']),
```
DoParkEQ, line 85: I think it'd be better if we rephrase `default` to `contained` to follow MUI's variant naming conventions.

src/chip.js
```diff
@@ -0,0 +1,95 @@
...
+    },
+  },
+  rectangle: {
+    borderRadius: '4px',
+  },
+}))
+
+const Chip = ({ clickable, color, onDelete, rectangle, variant, ...props }) => {
+  const theme = useTheme()
+  const styleProps = {
```
DoParkEQ, line 41: Can we allow users to set up custom background color?

src/chip.js
SamSverko: Using default is what is following MUI's convention for the [Chip component](https://material-ui.com/api/chip/#props):
Screen Shot 2020-09-24 at 3 57 45 PM

src/chip.js
SamSverko: We can, but I don't think it's necessary. We don't allow that for the `Button` or `Alert` components. We should keep things consistent for now. Looks like MUI keeps it to `primary`, `secondary`, `success`, `error`, `warning`, and `info`. That seems to cover our bases.

DoParkEQ: Oh, it is because we have the use case for the marketplace where `chip` has a different color than those to distinguish item cards.

stories/chip.stories.js
SamSverko: In Storybook controls, it defaults to look inside your `/src/chip` file to get the meta-data to populate, which is why we define the `PropTypes` as such:
```jsx
  /**
    * If true, the chip will appear clickable.
  */
  clickable: PropTypes.bool,
```

It is only when we do not do anything custom with a default MUI prop, such as `label` in this case. For that, we need to add the prop details to the Storybook story, because we do not define the `PropTypes` in the component.

Since the package code is in the `/src`, we should aim to document the `PropTypes` and descriptions inside each component. Then for any MUI-default props, we should keep that within their respective Story. This way we know anything inside our component is ours, anything inside a Story is not ours (i.e. MUI defaults).

src/chip.js
DoParkEQ: Oh I see, I am surprised that they didn't match this with their button component.

src/chip.js
DoParkEQ: Screen Shot 2020-09-24 at 4 07 01 PM

SamSverko: I'm looking at the Zeplin assets now, and I only see colours that should fall under the `primary`, `secondary`, `success` (green), `error`, `warning` (yellow), and `info` (blue/purple) list.

SamSverko: Exactly, that image should have chips where their variant is success (green), warning (yellow), and info (purple).

(I'm not sure why we are using info as purple, and not blue...

DoParkEQ: It is my fault that I didn't specify that the chip can have different colors than the system, primary and secondary colors.

SamSverko: okay, I can work that in. No problem :)

stories/chip.stories.js
SamSverko: Using `argTypes` in Stories is basically the "fallback" you use if your component itself cannot provide the proper documentation to the Story controls.

If we write our components `PropTypes` properly, we need not use argTypes.

src/chip.js
DoParkEQ: The colors you see in the image are different than system colors defined in `palette.js`. (Though it looks similar) They are not the colors showing the status of the item but distinguishing one from another like labels.

SamSverko: Ah, gotcha.

stories/chip.stories.js
DoParkEQ: I see. If I understood your point of view correctly, we are using both PropTypes and ArgTypes to separate MUI and custom props. (Correct me if it is inaccurate) Then should we validate our custom props through PropTypes inside of the `chip.stories.js` instead? I see that `onDelete` is included in PropTypes in `chip.js`

My level of knowledge about Storybook 6.0 is rudimentary but I thought PropTypes based Arg table generation is a fallback feature. Mainly because Storybook itself is functionally complete without PropTypes.

I see `argTypes` as a place where an implementor can see the entire list of props and its variants at once. That said, my personal struggle has been jumping two places back and forth to identify what props would a component take. If you have any ideas/suggestions for this specific problem, let me know!

SamSverko: So we want to document props originated from MUI in .stories.js and custom props in the component's JS file. Did I understand it correctly?

That is correct.
I do agree that we should be centralizing the `propTypes` info, and that it _should_ be under the `/src/[component]` because that is the code that ships. However, we have to balance between `Stories` and `src` because Storybook is still ironing out it's documentation. You can see that I opened an [issue on the official storybook repo](https://github.com/storybookjs/storybook/issues/11983) with a bug that has yet to be fixed. Looks like it has been recently labeled as `P1`, so hopefully it gets fixed soon!

I think for the time being we should just put everything inside the `src` side, and then all the remaining props under Storybook.

DoParkEQ (approved):

woozyking (approved):

PR #81, opened 2020-07-30
[G2M] StyleCardContainer - props/internal logic update

DoParkEQ:


hyx131:
looks good, except there are conflicts with master, rebase and push again before merge

PR #83, opened 2020-08-06
[G2M]table - update when sort changes dynamically [build]

tlowande:
remember still works for table toggle, not for the button chosen though

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.7.2] - 2020-08-07
+### Fixed
+- `<Table />` - table can now update based on dinamically sortBy
+
 ## [1.7.0] - 2020-07-22
 ### Added
 - `<ImageSlider>` - - LinkTo prop added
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.7.2-alpha] - 2020-08-06
```
woozyking, line 9: Keep alpha ones under unreleased. Released are only for non-alpha versions

stories/table.stories.js
```diff
@@ -286,3 +286,25 @@ export const rememberSortBy = () => {
     </>
   )
 }
+export const dynamicSortBy = () => {
+  const [sort, setSort] = useState('province')
+  const remember = {
```
woozyking, line 291: Since remember doesn’t work, rather remove it altogether and keep the scope of this story “unit”

tlowande: remember still works for table if you use the table toggle though, doesn't work for the values passed dynamically thorugh button. But ye let's keep it clean

CHANGELOG.md
tlowande: not sure what to do here since I made a non-alpha release already (by mistake) with an alpha tag. Since it's released though, I kept under `1.7.1` only

PR #85, opened 2020-08-10
[G2M] Eslint/standardize

SamSverko:
There has been some talk that our codebase formatting isn't as standardized as it could be. So let's take this PR to discuss and enforce an appropriate ESLint configuration for this project.

Here are the current "rules" the ESLint is enforcing:

| Rule        | Enforcement           | Details  |
| --------- | ------------------ | -------- |
| [comma-dangle](https://eslint.org/docs/rules/comma-dangle) | `'always-multiline'` | requires trailing commas when the last element or property is in a different line than the closing `]` or `}` and disallows trailing commas when the last element or property is on the same line as the closing `]` or `}` |
| [eol-last](https://eslint.org/docs/rules/eol-last) | `'always'` | enforces that files end with a newline (LF) |
| [indent](https://eslint.org/docs/rules/indent) | `'2-space'` | enforce consistent 2-space indentation |
| [object-curly-spacing](https://eslint.org/docs/rules/object-curly-spacing) | `'always'` | requires spacing inside of braces (except `{}`) |
| [quotes](https://eslint.org/docs/rules/quotes) | `'single', { 'avoidEscape': true }` | allows strings to use single-quotes or double-quotes so long as the string contains a quote that would have to be escaped otherwise |
| [semi](https://eslint.org/docs/rules/semi) | `'never'` | disallows semicolons as the end of statements (except to disambiguate statements beginning with `[`, `(`, `/`, `+`, or `-`) |

```
// comma-dangle
var foo = {
    bar: 'baz',
    qux: 'quux'
} // good

var foo = {
    bar: 'baz',
    qux: 'quux',
} // bad

// object-curly-spacing
var obj = { foo: 'bar' } // good
var obj = {foo: 'bar'} // bad
```

These rules were put together as starting ones, we can easily change them if the team requires it. All files in this PR have been formatted to the above rules so you can get a sense of what they look like in action.

Let's now discuss how we want to move forward with the ESLint configuration. Are there are rules missing/need to be changed? You can find the list of [ESLint rules here](https://eslint.org/docs/rules/).


.eslintrc.js
```diff
@@ -8,28 +8,28 @@ module.exports = {
   settings: {
     react: {
       pragma: 'React',
-      version: '16.13',
-    },
+      version: '16.13'
```
woozyking, line 11: we can use `version: 'detect'` to make this relatively more future-proof

.eslintrc.js
```diff
@@ -8,28 +8,28 @@ module.exports = {
...
     ecmaVersion: 2018,
-    sourceType: 'module',
+    sourceType: 'module'
   },
-  plugins: [
-    'react',
-  ],
+  plugins: ['react'],
   rules: {
+    'comma-dangle': ['error', 'never'],
```
woozyking, line 28: i would encourage the use of `comma-dangle` actually, for the very reason articulated in https://eslint.org/docs/rules/comma-dangle (it highlights better diffs for multiple lines)

specifically, i would encourage the use of `['error', 'always-multiline']` to ensure this advantage of using it is enforced at where it matters only.

.eslintrc.js
```diff
@@ -8,28 +8,28 @@ module.exports = {
...
+    'eol-last': ['error', 'always'],
     indent: ['error', 2],
-  },
+    'object-curly-spacing': ['error', 'always'],
+    quotes: ['error', 'single'],
```
woozyking, line 32: i would suggest the use of `avoidEscape` option here to allow the use of double quotes when single quotes literals are needed in the string, as articulated here https://eslint.org/docs/rules/quotes#avoidescape

.eslintrc.js
SamSverko: Oooh, I like that multiline approach!

.eslintrc.js
SamSverko: Oh, smart. Keeps it clean should we need to use both single and double in one string.

PR #86, opened 2020-08-11
[G2M]filter - filter component with checkbox select options

hyx131:


stories/filter.stories.js
```diff
@@ -0,0 +1,115 @@
...
+
+export default {
+  component: Filter,
+  title: 'Filter',
+}
+
+export const Default = () => {
+  const options = ['hello', 'there']
+  const [open, setOpen] = useState(false)
+  const [filterVals, setFilterVals] = useState({
```
DoParkEQ, line 16: What do you think about having both `options` and `filterVals` altogether in a single variable instead of separating them? In this way, we can reduce the number of props being passed. For example:
```
const checkboxConfig = {
          listOptions: ['hello', 'there', 'how', 'are', 'you'],
          currentOptions: ['hello', 'there'],
}
```

src/filter.js
```diff
@@ -0,0 +1,95 @@
...
+                      />
+                    )}
+                    label={c}
+                  />
+                </div>
+              ))}
+            </FormGroup>
+          </FormControl>
+          {selectAllOnClick && <div className={classes.select}>
+            <DynamicButton
```
DoParkEQ, line 63: I suggest modularizing grouped checkbox (however you name it) and `<DynamicButton/>` to make `<Filter/>` component more flexible in terms of reusability. Therefore filter becomes a container component
```
Filter ----- GroupedCheckboxes
         \_____ DynamicButton
```

src/filter.js
```diff
@@ -0,0 +1,95 @@
...
+      <Paper>
+        <Grid container justify='flex-start' alignItems='flex-start' direction='column'>
+          <FormControl component='fieldset' className={classes.formControl}>
+            <FormLabel component='legend' className={classes.formLabel}>{formLabel}</FormLabel>
+            <FormGroup>
+              {options.map((c) => (
+                <div key={c}>
+                  <FormControlLabel
+                    control={(
+                      <Checkbox
```
DoParkEQ, line 47: It should be replaced with `<StyledCheckbox/>`

src/filter.js
```diff
@@ -0,0 +1,95 @@
...
+
+Filter.propTypes = {
+  options: PropTypes.array.isRequired,
+  open: PropTypes.bool.isRequired,
+  checkboxOnChange: PropTypes.func.isRequired,
+  filterVals: PropTypes.object.isRequired,
+  anchorEl: PropTypes.any,
+  selectAllOnClick: PropTypes.func,
+  selectedAll: PropTypes.bool,
+  formLabel: PropTypes.string,
```
DoParkEQ, line 85: `formLabel` should be always required

src/filter.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+Filter.defaultProps = {
+  anchorEl: '',
+  selectAllOnClick: null,
+  selectedAll: false,
+  formLabel: '',
```
DoParkEQ, line 92: We should prevent the case where labels are missing for checkboxes (and other selection controls)

src/filter.js
```diff
@@ -0,0 +1,95 @@
...
+              ))}
+            </FormGroup>
+          </FormControl>
+          {selectAllOnClick && <div className={classes.select}>
+            <DynamicButton
+              type='tertiary'
+              style={{ padding: '0px', margin: '0px 0px 15px 0px' }}
+              onClick={selectAllOnClick}
+            >
+              {selectedAll ? 'Unselect All' : 'Select All'}
```
DoParkEQ, line 68: Buttons says `Select all` when all checkboxes are checked.
Screen Shot 2020-08-11 at 6 28 17 PM

DoParkEQ: For the style of `Categories` label: 
1) Its color should be black. 
2) Its color should be static. It changes its color to grey when clicked.
3) Its font type should be `subtitle`

stories/filter.stories.js
```diff
@@ -0,0 +1,115 @@
...
+  const [open, setOpen] = useState(false)
+  const [selectAll, setSelectAll] = useState(false)
+  const [filterVals, setFilterVals] = useState({
+    hello: false,
+    there: false
+  })
+  const ref = useRef(null)
+
+  const handleSelectAll = () => {
+    const selectedVals = options.reduce((obj, opt) => ({ ...obj, [opt]: !selectAll }), {})
```
DoParkEQ, line 87: Since all values of each element changes based on `selectAll` I think it can be:
```
const len = options.length;
const selectedVals = new Array(len).fill(selectedAll);
setFilterVals(selectedVals);
```

stories/filter.stories.js
```diff
@@ -0,0 +1,115 @@
...
+      />
+    </Grid>
+  )
+}
+
+export const SelectAll = () => {
+  const options = ['hello', 'there']
+  const [open, setOpen] = useState(false)
+  const [selectAll, setSelectAll] = useState(false)
+  const [filterVals, setFilterVals] = useState({
```
DoParkEQ, line 80: What do you think about converting `filterVal` to object in array? In this way, we can directly use `.map()` to render checkbox based on its status without `options`.

src/filter.js
DoParkEQ: For the style of Categories label:

Its color should be black.
Its color should be static. It changes its color to grey when clicked.
Its font type should be subtitle

stories/filter.stories.js
hyx131: yep, sounds good

src/filter.js
hyx131: oh oki that makes sense, would there be any cases where components other than button is used to open the filter select popper? (if so should we make that as a placeholder and pass the button down from the parent to occupy that placeholder instead?)

src/filter.js
hyx131: oki

src/filter.js
hyx131: think the color change results form <FormLabel>'s theme itself, will look into ways to override that

stories/filter.stories.js
hyx131: yep that sounds good

stories/filter.stories.js
hyx131: used the object in array method

DoParkEQ:
Nice work! I have a couple of discussion points regarding the recent change.

**Component Hierarchy**
It is my fault that I did not update the diagram after our discussion. I imagined the component hierarchy to be more like this: 

Screen Shot 2020-08-24 at 2 01 21 PM

To me, the current `SelectionGroup` is the original `Filter` with a different name. In this diagram, all selection control types are wrapped under `SelectionGroup`. The difference with the current implementation would be: 

1. Grouping all four types of controls will be covered under SelectionGroup. It means that we don't need to adopt `<Type>Group` for each control type.
2. SelectionGroup itself is detached from other components such as `Grid`, `FormControl`, `Typography`, `DynamicButton` and etc. 

**Data type**
Right now we options accept three data types: object, string, and number(integer and float). I noticed that if I want to switch control type from one to another, I also need to convert data types manually then feed them as `options` prop. 

E.g. I need to manually change my data type from `['option 1', 'option 2', 'option 3' ]` to `[{option 1: false},{option 2: false},{option 3: false}]` in order to change control type from radio to switch.

I feel that we should enforce a single data type (e.g. string) for all types of selection controls so that switching from one to another is more convenient from an end-user perspective.

src/selection-group/index.js
```diff
@@ -0,0 +1,128 @@
...
+  onChange,
+  direction,
+  hasSelectAll,
+  sliderStep,
+}) => {
+  const classes = useStyles()
+  const [filterVals, setFilterVals] = useState(options)
+  const [selectAll, setSelectAll] = useState(true)
+
+  const checkboxOnChange = (e) => {
```
DoParkEQ, line 36: I think the name of this should be changed as it is used for both `CheckboxGroup` and `SwitchGroup`

src/selection-group/index.js
```diff
@@ -0,0 +1,128 @@
...
+import Typography from '@material-ui/core/Typography'
+
+import DynamicButton from '../dynamic-button'
+import CheckboxGroup from './checkbox-group'
+import RadioGroup from './radio-group'
+import SwitchGroup from './switch-group'
+import SliderGroup from './slider-group'
+
+
+const useStyles = makeStyles(() => ({
```
DoParkEQ, line 17: I recommend using theme.spacing prop for margins/paddings or at least have values that are a multiplication of 4.

hyx131:
> Nice work! I have a couple of discussion points regarding the recent change.
> 
> **Component Hierarchy**
> It is my fault that I did not update the diagram after our discussion. I imagined the component hierarchy to be more like this:
> 
> Screen Shot 2020-08-24 at 2 01 21 PM
> 
> To me, the current `SelectionGroup` is the original `Filter` with a different name. In this diagram, all selection control types are wrapped under `SelectionGroup`. The difference with the current implementation would be:
> 
> 1. Grouping all four types of controls is covered under SelectionGroup. It means that we don't need to adopt `<Type>Group` for each control type.
> 2. SelectionGroup itself is detached from other components such as `Grid`, `FormControl`, `Typography`, `DynamicButton` (just like `CheckboxGroup`)
> 
> **Data type**
> Right now we options accept three data types: object, string, and number(integer and float). I noticed that if I want to switch control type from one to another, I also need to convert data types manually then feed them as `options` prop.
> 
> E.g. I need to manually change my data type from `['option 1', 'option 2', 'option 3' ]` to `[{option 1: false},{option 2: false},{option 3: false}]` in order to change control type from radio to switch.
> 
> I feel that we should enforce a single data type (e.g. string) for all types of selection controls so that switching from one to another is more convenient from an end-user perspective.

Thanks, will review!

PR #88, opened 2020-08-12
[G2M] dependency check job for master workflow, include node 14.x into CI matrix

woozyking:


.github/workflows/master.yml
```diff
@@ -92,3 +92,20 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set Node.js version
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - run: npm i -g depcheck
```
SamSverko, line 110: Could we consolidate into `npx depcheck`? Mostly just curious about using `npx`.

woozyking: yea `npx` is fine, but it might not leverage the built-in "global" node_modules cache. but i like running things fresh, let's do that then

PR #92, opened 2020-08-17
[G2M] Storybook - Upgrade deps and add controls to components

SamSverko:
- Upgrade Storybook to version 6.
- Adding Storybook controls to our component stories for better documentation and demonstration.


src/table/table-toolbar/download.js
```diff
@@ -10,7 +10,7 @@ import ClickAwayListener from '@material-ui/core/ClickAwayListener'
 import Badge from '@material-ui/core/Badge'
 import SaveAltIcon from '@material-ui/icons/SaveAlt'
 
-import { DynamicButton } from '../../index'
```
DoParkEQ, line 13: Does this change still accommodate the projects that use `<DynamicButton/>`? I remember that there was a certain procedure for component deprecation.

stories/alert.stories.js
```diff
@@ -1,76 +1,82 @@
...
-  ))
-  .add('Success dif icon', () => (
-    <Alert
-      iconMapping={{ success: <VisibilityOff fontSize="inherit" /> }}
-      severity="success"
-      message="Success message."
-    />
-  ))
+export default {
+  title: 'Lab/Alert',
```
DoParkEQ, line 7: Which components goes under `Lab`?

src/table/table-toolbar/download.js
SamSverko: That's a good question. I'm not too sure about that. I can re-add DynamicButton as a component, that way we have both DynamicButton and Button, and projects can transition over to Button over time.

stories/alert.stories.js
SamSverko: I was just mirroring [MUI-categorization](https://material-ui.com/). You can see that `Alert` is under `Lab`. We can move things into any category we want.

Screen Shot 2020-08-20 at 4 36 57 PM

src/table/table-toolbar/download.js
DoParkEQ: I think it'd be safe to have DynamicButton under `Deprecated` until we replace all old buttons to a new one.

stories/alert.stories.js
DoParkEQ: I see, I think we can also use `Lab` for a temporary home for individual components that are not yet grouped with others.

DoParkEQ:
Seems good! 👍  I noticed is that Storybook casts `<Loader/>` on the entire `Docs` page. Is there any way to make the component only visible in a window of each section?
Screen Shot 2020-08-24 at 12 06 08 PM

SamSverko:
I feel like that is a bug on Storybook's side, @DoParkEQ . I've noticed Storybook put out a log of minor patches recently to address a lot of bugs, this may just be another bug for Storybook to squash.
Maybe we can create an issue for this to keep it on our radar?

DoParkEQ:
Sure, let's do that

stories/widget-stats.stories.js
```diff
@@ -1,68 +1,29 @@
...
 
-storiesOf('WidgetStats', module)
-  .add('Default', () => <WidgetStats />)
-  .add('Number', () => (
-    <WidgetStats
-      title='Example Number'
-      value={parseInt((Math.random() * (100 - 1) + 1).toFixed(0))}
-    />
-  ))
-  .add('Trend', () => (
```
woozyking, line 14: the number/trend stories were used to demonstrate how `WidgetStats` is used to deprecate both `WidgetNumber` and `WidgetTrend` components.

i'd suggest keeping these two in separate [named CSF](https://storybook.js.org/docs/react/api/csf#named-story-exports) examples for that (we can drop them though after removing the above mentioned deprecated components fully in next major version release)

SamSverko: So create 2 new stories just for trend?

SamSverko: Im confused on the comment here.

SamSverko: Do you mean uncommenting these lines in `/src/index.js`?
```js
// export { default as WidgetNumber } from './widget-number' // deprecated by WidgetStats
// export { default as WidgetTrend } from './widget-trend' // deprecated by WidgetStats
```

woozyking: below `export default`, add two more named exports

```jsx
export const number = () => ...
export const trend = () => ...
```

SamSverko: But we have the controls to preview all the states of the component...?

woozyking: always need to keep a good balance between pure abstractions and some materializations. giving a power user/developer the flexibility of controls to permute through various use cases by themself is nice, but for someone without the explicit context of what the components are capable of can still benefit from having some prepared examples.

SamSverko: Gotcha, that makes sense. I'll add those variations in.

PR #94, opened 2020-08-17
[G2M] Widget stat/handle zero divide

### LINKED ISSUE #9: WidgetStat - handle 0/0 calculation
currently will produce NaN if 0s are passed

SamSverko:
Fix #9

### DIFF
--- a/src/styled-card-container.js
+++ b/src/styled-card-container.js
@@ -79,7 +79,7 @@ const useStyles = makeStyles((theme) => {
 })
 
 const StyledCardContainer = ({ pattern, onClick, selected, children, ...rest }) => {
-  const whichStyle = 'style' + pattern.style
+  const whichStyle = `style${pattern.style}`
   const classes = useStyles(pattern)
   
   return (
--- a/src/styled-card-container.js
+++ b/src/styled-card-container.js
@@ -78,9 +78,9 @@ const useStyles = makeStyles((theme) => {
   }
 })
 
-const StyledCardContainer = ({ pattern, onClick, selected, children, ...rest}) => {
-  const whichStyle = "style" + pattern.style;
-  const classes = useStyles(pattern);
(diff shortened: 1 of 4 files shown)


src/styled-card-container.js
```diff
@@ -78,9 +78,9 @@ const useStyles = makeStyles((theme) => {
   }
 })
 
-const StyledCardContainer = ({ pattern, onClick, selected, children, ...rest}) => {
-  const whichStyle = "style" + pattern.style;
-  const classes = useStyles(pattern);
+const StyledCardContainer = ({ pattern, onClick, selected, children, ...rest }) => {
+  const whichStyle = 'style' + pattern.style
```
woozyking, line 82:
```js
const whichStyle = `style${pattern.style}`
```

src/widget-stats.js
```diff
@@ -23,22 +23,28 @@ const WidgetStats = ({ title, value, prev, units, children, trendInfo }) => {
...
-  const trend = isTrendPercentage
-    ? Math.round(((value - prev) / prev) * 100).toLocaleString()
-    : Math.round(value - prev).toLocaleString()
+  let trend = isTrendPercentage
+    ? Math.round(((value - prev) / prev) * 100)
+    : Math.round(value - prev)
+
+  if (isNaN(trend)) {
+    trend = 0
+  }
```
woozyking, line 32: we can leverage the fact that `NaN || 0` yields `0`. also removing `toLocaleString` would alter the current behavior of the component and introduce a breaking change.

perhaps:

```js
const trend = ((isTrendPercentage
  ? Math.round(((value - prev) / prev) * 100)
  : Math.round(value - prev)) || 0).toLocaleString()
```

SamSverko: That looks much better to me, thank you!

Issue #95, opened 2020-08-20
[REQUEST] Add color palettes examples


DoParkEQ:
## Is your feature request related to a problem?

![Screen Shot 2020-08-20 at 5 07 00 PM](https://user-images.githubusercontent.com/59836949/90825895-95757300-e307-11ea-96bf-562a9258ad37.png)

`Colors` is one of the design elements for our UI. Right now we only have examples for `Typography` and `Shadows` in Storybook.

---

## Describe the solution you'd like

I'd like to add color palettes examples under `System`

---

## Are you able to assist bring the feature to reality?

Yes. I'll implement this myself to get familiar with storybook 6.0 and added features/recent changes in react-labs

SamSverko:
Love this!!

PR #103, opened 2020-08-27
[G2M] Storybook color palette example added

DoParkEQ:
Addresses #95 

- updated folder title from System to Theme
- created color palette examples.

### DIFF
--- a/stories/colors.stories.js
+++ b/stories/colors.stories.js
@@ -1,10 +1,14 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { makeStyles, useTheme } from '@material-ui/core/styles'
 import { getContrastRatio } from '@material-ui/core/styles/colorManipulator'
-import Box from '@material-ui/core/Box'
-import Typography from '@material-ui/core/Typography'
-
-const useStyles = makeStyles(theme => ({
+import Button from '../src/button'
+import { Box, Tooltip, Typography } from '@material-ui/core'
+const useStyles = makeStyles((theme) => ({
+  buttonContainer: {
+    display: 'flex',
+    justifyContent: 'flex-end',
+    marginRight: theme.spacing(2),
+  },
   container: {
     display: 'flex',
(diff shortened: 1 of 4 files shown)


stories/colors.stories.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  return (
+    <>
+      {renderPalettes(primaryColors, 'primary')}
+      {renderPalettes(secondaryColors, 'secondary')}
+      {renderPalettes(systemColors, 'system')}
+    </>
+  )
+}
```
SamSverko, line 76: These tiles look great!

If we want to get fancy, we can add a "copy hex value to clipboard" on tile select. Not necessary, but could be a useful way to quickly grab values when needed.

```suggestion
    const copyHexToClipboard = (event) => {
      if (!event.target.firstChild.textContent.includes('#')) {
        return
      }
      console.log(`Color selected: ${event.target.firstChild.textContent}`)
      const textarea = document.createElement('textarea')
      textarea.id = 'temp_element'
      textarea.style.height = 0
      document.body.appendChild(textarea)
      textarea.value = event.target.firstChild.textContent
      const selector = document.querySelector('#temp_element')
      selector.select()
      document.execCommand('copy')
      document.body.removeChild(textarea)
    }

    return (
      <div onClick={copyHexToClipboard}>
        <Typography variant='subtitle1' gutterBottom>{title}</Typography>
        <Typography variant='body1' style={{ fontWeight: 600, color: theme.palette.secondary[600] }} gutterBottom>{`theme.palette.${title}`}</Typography>
        <div className={classes.container}>
          {palettes}
        </div>
      </div>
    )
  }

  return (
    <>
      {renderPalettes(primaryColors, 'primary')}
      {renderPalettes(secondaryColors, 'secondary')}
      {renderPalettes(systemColors, 'system')}
    </>
  )
}
```

DoParkEQ: Like the idea! I added a copy to clipboard and toggle between hex code and shade label display.

PR #108, opened 2020-09-03
[G2M] Typography

### LINKED ISSUE #106: [REQUEST] abstract Typography component
## Is your feature request related to a problem?

The MUI Typography adds a bottom margin with `gutterBottom` prop. However, it does not offer an option to customize the height of the margin.

---

## Describe the solution you'd like

add a more flexible way to adjust its spacing with numerical values. 

---

## Are you able to assist bring the feature to reality?

Yes 

---

DoParkEQ:
fixes #106

### DIFF
--- a/stories/typography.stories.js
+++ b/stories/typography.stories.js
@@ -1,6 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import Typography from '@material-ui/core/Typography'
+import Typography from '../src/typography'
+import { Divider } from '@material-ui/core'
+import { useTheme } from '@material-ui/core/styles'
 
 export default {
   title: 'Theme/Typography',
@@ -8,6 +10,7 @@ export default {
   args: {
     children: 'Text',
     variant: 'h1',
+    marginBottom: 0,
   },
   argTypes: {
     children: {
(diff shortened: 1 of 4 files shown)


stories/typography.stories.js
```diff
@@ -1,13 +1,16 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import Typography from '@material-ui/core/Typography'
+import Typography from '../src/typography'
+import { Divider } from '@material-ui/core'
+import theme from '../src/theme'
```
SamSverko, line 5:
```suggestion
import { useTheme } from '@material-ui/core/styles'
```

stories/typography.stories.js
```diff
@@ -38,24 +41,82 @@ export default {
...
+  'subtitle1',
+  'body1',
+  'button',
+  'caption',
+  'overline',
+]
 
-const Template = ({ children, variant }) => <Typography variant={variant}>{children}</Typography>
+const Template = ({ variant, children, marginBottom }) => {
+
```
SamSverko, line 76:
```suggestion
  const theme = useTheme()
```

stories/typography.stories.js
SamSverko: By leveraging the `useTheme` hook, we don't need to re-import our theme twice (once in the global `ThemeProvider`, then another time with a manual `import theme from '...'`.

SamSverko (changes requested):
This is great! Really love the "playground" section of the `Typography` story.

Since we are creating our own `Typography` component, we should export it from `./src/index.js`, so that end-users can easily import it in their projects:

`./src/index.js`

```js
import Typography from './typography'
// ...
export {
  // ...
  Typography,
}

```

woozyking:
> This is great! Really love the "playground" section of the `Typography` story.
> 
> Since we are creating our own `Typography` component, we should export it from `./src/index.js`, so that end-users can easily import it in their projects:
> 
> `./src/index.js`
> 
> ```js
> import Typography from './typography'
> // ...
> export {
>   // ...
>   Typography,
> }
> ```

can just be `export { default as Typography } from './typography'`

SamSverko (approved):

PR #112, opened 2020-09-08
[G2M] Tests/button

SamSverko:
Wrote some tests for the button component.


src/button.js
```diff
@@ -15,13 +15,14 @@ const useStyles = makeStyles((theme) => {
...
 })
 
-const ButtonComponent = ({ children, isLoading, type, noSpacing, ...props }) => {
+const Button = ({ children, isLoading, noSpacing, type, ...props }) => {
   const classes = useStyles()
 
   const styleProps = {
     color: 'primary',
-    type: 'primary',
+    type: type,
```
woozyking, line 23: shorthand can be used here

```js
{
  color: 'primary',
  type,
}
```

__tests__/button/button.test.js
```diff
@@ -0,0 +1,97 @@
...
+    // not disabled
+    expect(button).not.toBeDisabled()
+    // medium size
+    expect(button).not.toHaveClass('MuiButton-sizeSmall', 'MuiButton-sizeLarge')
+    // primary type
+    expect(button).toHaveAttribute('type', 'primary')
+  })
+})
+
+describe('disabled state', () => {
```
woozyking, line 25: no need to test props that are relayed from MUI's Button component, it should also be removed from our Button's propTypes definitions

__tests__/button/button.test.js
```diff
@@ -0,0 +1,97 @@
...
+    const { getByTestId } = render(<Button size='medium'>My Button</Button>)
+
+    expect(getByTestId('button')).not.toHaveClass('MuiButton-sizeSmall', 'MuiButton-sizeLarge')
+  })
+
+  it('should be large', () => {
+    const { getByTestId } = render(<Button size='large'>My Button</Button>)
+
+    expect(getByTestId('button')).toHaveClass('MuiButton-sizeLarge')
+  })
```
woozyking, line 79: testing props permutations can be done through `test.each` interface to emulate what's commonly known as "test matrix" (similar concept can be seen in github actions node versions matrix, etc.)

```jsx
it.each([
  ['small', ['MuiButton-sizeSmall']],
  ['medium', ['MuiButton-sizeSmall', 'MuiButton-sizeLarge']],
  // ... more cases
])('%p, %p', (size, classes) => {
  const { getByTestId } = render(<Button size={size}>My Button</Button>) 
  expect(getByTestId('button')).toHaveClass(...classes)
}
```

same can be applied to almost all other props testing

SamSverko: Oh, that's interesting! Let me try that out.

SamSverko:
@woozyking , I realized the `depcheck` fails on `@testing-library/jest-dom` because it's not referenced in any file other than the` package.json`.

In the `package.json`, it's located here:
```js
  "jest": {
    "setupFilesAfterEnv": [
      "@testing-library/jest-dom/extend-expect"
    ]
  },
```

Is there a way we can add an ignore flag to our `depcheck` check?

woozyking (approved):

PR #116, opened 2020-09-10
[G2M] Package - Update to v1.8.1.

SamSverko:
To update package to `v1.8.1`.


package.json
```diff
@@ -72,4 +72,4 @@
       "limit": "100 KB"
     }
   ]
-}
+}
```
woozyking, line 75: maybe keep the empty newline to be more "unix-compliant"

SamSverko: Good catch! Thanks.

PR #121, opened 2020-09-15
[G2M] Themeprovider/export

SamSverko:
Update to export our own `ThemeProvider` that uses the `react-labs` theme by default. Instead of requiring `@material-ui/core` as well as importing the `react-labs` theme.

Before:
```jsx
import React from "react";
import ReactDOM from "react-dom";
import { ThemeProvider } from "@material-ui/core";
import { Theme } from "@eqworks/react-labs";

import App from "./App";

ReactDOM.render(
  <ThemeProvider theme={Theme}>
    <App />
  </ThemeProvider>,
  document.getElementById("root")
);
```

After:
```jsx
import React from "react";
import ReactDOM from "react-dom";
import { ThemeProvider } from "@eqworks/react-labs";

import App from "./App";

ReactDOM.render(
  <ThemeProvider>
    <App />
  </ThemeProvider>,
  document.getElementById("root")
);
```

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,10 +1,10 @@
 import React from 'react'
 
-import { Theme, ThemeProvider } from '../src/index'
+import { ThemeProvider } from '../src/index'
 
 export const decorators = [
   (Story) => (
-    <ThemeProvider theme={Theme}>
+    <ThemeProvider>
       <Story />
     </ThemeProvider>
   ),
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,7 +1,6 @@
 import React from 'react'
-import { ThemeProvider } from '@material-ui/core/styles'
 
-import { Theme } from '../src/index'
+import { Theme, ThemeProvider } from '../src/index'
 
 export const decorators = [
   (Story) => (
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,22 @@
 // theme
-export { ThemeProvider } from '@material-ui/core/styles'
-export { default as Theme } from './theme/index'
+import React from 'react'
+import PropTypes from 'prop-types'
+import { ThemeProvider as MUIThemeProvider } from '@material-ui/core/styles'
+import DefaultTheme from './theme'
+
+export { DefaultTheme }
+export const ThemeProvider = ({ children, theme = DefaultTheme }) => {
+  return (
+    <MUIThemeProvider theme={theme}>
+      {children}
+    </MUIThemeProvider>
+  )
+}
+
+ThemeProvider.propTypes = {
+  children: PropTypes.node,
+  theme: PropTypes.object,
+}
 
 // components
 export { default as Alert } from './alert'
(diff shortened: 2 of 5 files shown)


.storybook/preview.js
```diff
@@ -1,11 +1,10 @@
 import React from 'react'
-import { ThemeProvider } from '@material-ui/core/styles'
 
-import { Theme } from '../src/index'
+import { ThemeProvider } from '../src/index'
```
DoParkEQ, line 3: Can we also support default export on this one?

SamSverko: What do you mean by "default"? Are you referring to `ThemeProvider` or `DefaultTheme`?

Right now, if you do not supply a `theme` prop to the `ThemeProvider`, it will use the `DefaultTheme` (our `react-labs` theme) as a fallback. However, anyone can create their own theme, and pass it as a prop to the `ThemeProvider`.

```
import { ThemeProvider } from '../src/index'

export const decorators = [
  (Story) => (
    <ThemeProvider>
      <Story />
    </ThemeProvider>
  ),
]
```

is the same as:

```
import { DefaultTheme, ThemeProvider } from '../src/index'

export const decorators = [
  (Story) => (
    <ThemeProvider theme={DefaultTheme}>
      <Story />
    </ThemeProvider>
  ),
]
```

<hr />

I've also exported our `DefaultTheme` object should we need to use it anywhere else.

`Inside /src/index:`
```
export { DefaultTheme }
// this allows us to use this anywhere:
// import { DefaulTheme } from '@eqworks/react-labs'
```

src/index.js
```diff
@@ -1,5 +1,22 @@
...
+    <MUIThemeProvider theme={theme}>
+      {children}
+    </MUIThemeProvider>
+  )
+}
+
+ThemeProvider.propTypes = {
+  children: PropTypes.node,
+  theme: PropTypes.object,
+}
```
woozyking, line 19: we can keep this entire scope in `./theme/`, so `src/index.js` can export it as:

```js
export { default as Theme, ThemeProvider } from './theme'
```

and inside `./theme/index.js` (or another module within `./theme/` directory if preferred), host the implementation of this theme provider wrapper. which would allow the unit import as @DoParkEQ commented in https://github.com/EQWorks/react-labs/pull/121/files#r488870184

```js
import { ThemeProvider } from '@eqworks/react-labs/dist/theme`
```

effectively bypassing the import of the entire root index (which would involve the need of installing many peer dependencies that may not be used in a given application context)

SamSverko: Ooh, I like this implementation. Let me try it out :)

src/theme/index.js
```diff
@@ -1,16 +1,31 @@
...
 import { createMuiTheme } from '@material-ui/core/styles'
+
 import overrides from './overrides/index'
 import palette from './palette'
 import props from './props'
 import shadows from './shadows'
 import typography from './typography'
 
-const theme = createMuiTheme({
+export const DefaultTheme = createMuiTheme({
```
woozyking, line 12: you can also keep a `export default DefaultTheme` after this just so that this acts as a slight backward compatible measure, so we're not going to breach the by-book definition of API breaking change on this one.

woozyking (approved):

PR #122, opened 2020-09-17
[G2M] Tests/Alert

SamSverko:
Add tests for `<Alert />` component.

### DIFF
--- a/src/theme/index.js
+++ b/src/theme/index.js
@@ -30,3 +30,4 @@ ThemeProvider.propTypes = {
   theme: PropTypes.object,
 }
 
+export default DefaultTheme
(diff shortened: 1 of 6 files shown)


tlowande (approved):

src/theme/index.js
```diff
@@ -30,4 +30,6 @@ ThemeProvider.propTypes = {
   theme: PropTypes.object,
 }
 
-export default DefaultTheme
+export  {
+  DefaultTheme,
+}
```
woozyking, line 35: :thonk: i feel this looks familiar

can we keep both named and default export for this for backward compat?

SamSverko: Haha, yes I noticed that as well. I will make it the same as on the [refetch-control PR](https://github.com/EQWorks/react-labs/pull/123)

PR #123, opened 2020-09-21
[G2M] Refetch control

SamSverko:
I’m in the process of building out this “agnostic data fetcher UI” (think of the Gmail refresh button). Where it:

- :white_check_mark: Fetches your data using [react-query](https://react-query.tanstack.com/).
- :white_check_mark: Displays visual feedback (`loading` & `error` states).
- :x: Passes the data back to its parent component?

I’m not sure how to handle the last part. Any thoughts/help?


src/theme/index.js
```diff
@@ -30,4 +30,6 @@ ThemeProvider.propTypes = {
   theme: PropTypes.object,
 }
 
-export default DefaultTheme
+export {
+  DefaultTheme,
+}
```
woozyking, line 35: these two do different things. both should be kept.

```js
// this is the same as export { DefaultTheme }
// which is a named-export
export const DefaultTheme = ...
                                                     
// this is for backward compatibility
// so it doesn't break previous use as module default
export default DefaultTheme
```

stories/refetch-data.stories.js
```diff
@@ -0,0 +1,21 @@
...
+  args: {},
+  argTypes: {
+    update: {
+      control: null,
+    },
+  },
+}
+
+const Template = (args) => (
+  <RefetchData {...args} />
```
woozyking, line 17: perhaps provide a better demonstration with an actual `update` function for `onClick` which mutates a `status` state.

`setTimeout` can be used to emulate something that's happening that takes some time (such as performing an API call) so `loading` can be prominent for the emulation period. and `error` status can potentially be pseudo-randomly set by using `Math.random` to further emulate a realistic API call experience.

src/refetch-data.js
```diff
@@ -0,0 +1,90 @@
...
+          className={classes.alert}
+          message='Loading data...'
+          severity='info'
+          variant='standard'
+          width='100%'
+        />
+      )}
+      <div className={classes.container}>
+        <button className={classes.button} onClick={update}><RefreshIcon /></button>
+      </div>
```
woozyking, line 74: one of the important factors for the use of libraries such as `react-query` is the capability to emulate "real-time" updates, so it's quite useful (but not necessarily required) to provide a visual indicator for _when_ the data was last updated, such as accepting an `updated` prop which is in the form a JS Date object, which then can be formatted using `localeString` (or provided as a string to be displayed as-is)

woozyking (changes requested):

src/theme/index.js
SamSverko: Ohh, I see how that backward compatibility option is useful, thank you.

stories/refetch-data.stories.js
```diff
@@ -0,0 +1,64 @@
...
+      setFetchStatus('none')
+      setLastUpdated(new Date(Date.now()))
+      console.log(data)
+      return data
+    }, 1000)
+  }
+
+  if (status === 'error') {
+    setFetchStatus('error')
+  }
```
tlowande, line 34: I lost track of this variable. Where is `status` declared again or is it used for something??

src/theme/index.js
```diff
@@ -9,7 +9,7 @@ import props from './props'
 import shadows from './shadows'
 import typography from './typography'
 
-const DefaultTheme = createMuiTheme({
+export const DefaultTheme = createMuiTheme({
```
tlowande, line 12: the final export, on line 33, shouldn't it be for ThemeProvider?

src/refetch-data.js
```diff
@@ -0,0 +1,179 @@
...
+  */
+  status: PropTypes.oneOf(['error', 'loading', 'none']),
+  /**
+    * The request method.
+  */
+  fetchDataFunction: PropTypes.func,
+  /**
+    * The date and time of the most recent return of data.
+  */
+  lastUpdated: PropTypes.instanceOf(Date),
```
tlowande, line 176: I noticed that there's no tooltip on the first render. I guess this prop could be `required` so the button is always initialized with the last time it was updated (prob as soon as data was fetched for the first time). Or set a default value to just `now` if that makes sense. 

There's also no tooltip if `error` but I guess if you have the porp as `required` or `default` should be fine.

src/theme/index.js
SamSverko: Not in our case because we've previously imported the `theme` object directly from `/src/theme/index`. It's just to provide some backward compatibility.

Any future cases will use this: `import { ThemeProvider, Theme } from '@eqworks/react-labs'`.

stories/refetch-data.stories.js
SamSverko: Status is a prop for the [RefetchData](https://github.com/EQWorks/react-labs/blob/aecb4c8a2019c7d38aa2377047f6a79171df6fae/src/refetch-data.js) component.

SamSverko: The stories I wrote for the `RefetchData` component are purely for demonstration purposes. They simple "mock" retrieving data to showcase the `status` displays of `loading` and `error`.

src/refetch-data.js
SamSverko: There's no tooltip on first render because you need to activate the refetch button to then be able to see when it was last called. If it was never called (i.e. on first render without selecting it), the tooltip should not become visible on hover.

src/refetch-data.js
```diff
@@ -29,29 +28,98 @@ const useStyles = makeStyles((theme) => {
...
+      const startTime = lastUpdated
+      const currentTime = new Date(Date.now())
+      const timeDifference = currentTime.getTime() - startTime.getTime()
+      const resultInMinutes = Math.round(timeDifference / 60000)
+      if (resultInMinutes < 1) {
+        setLastUpdatedDisplay('Updated just now')
+      } else if (resultInMinutes < 60) {
+        setLastUpdatedDisplay(`Updated ${resultInMinutes} mins ago`)
+      } else {
+        setLastUpdatedDisplay('Updated more than an hour ago')
```
woozyking, line 115: it's actually better to just show when _exactly_ they were updated as the UI will not rerender these values if `lastUpdated` doesn't change, say, for an hour, yet the display would still show something like "Updated just now", which may be misleading.

so you can actually use something like [here](https://github.com/EQWorks/snoke/pull/829/files#diff-b3be9e89df5c01b0c87735a8886b6b97R19):

```js
const formatUpdated = (updated) => (updated.toDateString() === new Date().toDateString()
  ? updated.toLocaleTimeString()
  : updated.toLocaleString())
```

which keeps a decent balance for any time from "today" to be shortened by having the date portion omitted.

woozyking: another option, for if we want to keep the "ago" logic, is to attach an active effect through "new Date()" to force the re-render and let the "ago" display to be refreshed

stories/refetch-data.stories.js
```diff
@@ -7,15 +8,57 @@ export default {
...
+  const [fetchStatus, setFetchStatus] = useState('none')
+  const [lastUpdated, setLastUpdated] = useState(undefined)
 
-export const Default = Template.bind({})
+  const fetchData = async () => {
+    setFetchStatus('loading')
+    const { data } = await axios.get('https://jsonplaceholder.typicode.com/users/1')
+    window.setTimeout(() => {
+      setFetchStatus('none')
+      setLastUpdated(new Date(Date.now()))
```
woozyking, line 26: `new Date()` is actually equivalent to `new Date(Date.now())` by value

```js
new Date().toISOString() === new Date(Date.now()).toISOString() // -> true
```

stories/refetch-data.stories.js
woozyking: variable/reference `status` doesn't exist in this module (`refetch-data.stories.js`), nor this `Default` component

status

perhaps you meant to just use `fetchStatus` here, and `setFetchStatus('error')` in the HTTP get request?

```js
try {
  const { data } = axios.get(...)
} catch(error) {
  setFetchStatus('error')
}
```

stories/refetch-data.stories.js
```diff
@@ -7,15 +8,57 @@ export default {
...
-)
+export const Default = () => {
+  const [fetchStatus, setFetchStatus] = useState('none')
+  const [lastUpdated, setLastUpdated] = useState(undefined)
 
-export const Default = Template.bind({})
+  const fetchData = async () => {
+    setFetchStatus('loading')
+    const { data } = await axios.get('https://jsonplaceholder.typicode.com/users/1')
+    window.setTimeout(() => {
```
woozyking, line 24: if an actual remote HTTP API is involved, then `setTimeout` is probably not needed to emulate that.

also, why not make use of this data by displaying it in the story?

stories/refetch-data.stories.js
```diff
@@ -7,15 +8,57 @@ export default {
...
+      setFetchStatus('none')
+      setLastUpdated(new Date(Date.now()))
+      console.log(data)
+      return data
```
woozyking, line 28: what does this `return` do?

woozyking (changes requested):

src/refetch-data.js
```diff
@@ -0,0 +1,179 @@
...
+          onClose={handleClose}
+          onOpen={handleOpen}
+          open={open}
+          title={lastUpdatedDisplay}
+        >
+          <button
+            aria-label='Refresh data'
+            className={classes.button}
+            onKeyDown={handleKey}
+            onMouseUp={handleClick}
```
woozyking, line 153: should we disable the button when `status === 'loading'`? otherwise it'll perform as many times of `handleClick` as the button is clicked, which can lead to unexpected abuse.

stories/refetch-data.stories.js
SamSverko: I'm confused here. Doesn't it not matter how the fetch is implemented in the stories because they were mocked just to visualize the `RefetchData` UI updates?

`status` is just a prop of the `RefetchData` component. You can see it here inside `refetch-data.stories.js`:
```jsx
return <RefetchData
    status={fetchStatus}
    fetchDataFunction={fetchData}
    lastUpdated={lastUpdated}
  />
```

src/refetch-data.js
SamSverko: I did add an interval that occurs every minute to update the lastUpdated display value:
```jsx
const updateLastUpdatedTimer = setInterval(() => {
  displayLastUpdated()
}, 60 * 1000)
```

This works well. If you select the button and wait 1 minute, the lastUpdated display will update to "Updated 1 minute ago" etc...

stories/refetch-data.stories.js
SamSverko: If I use `new Date().toISOString()` instead of `new Date(Date.now()).toISOString()`, I get this error from the `RefetchData` component: 
```
startTime.getTime is not a function
```

stories/refetch-data.stories.js
tlowande: ye, it does not matter how you're fetching your data, but you're not handling the error here. 

so, inside this story function you built, `status` doesn't exist. `status` is what you're passing to the `<RefetchData>` component, in this case, you named it `fetchStatus`. The story (the outer function that mounts `<RefetchData>`) only knows `fetchStatus` as a variable.

so just to simulate the error condition, you'd need a `try/catch` block, to `setFetchStatus('error')` so your component will handle it as it should.

or just remove this `if` block since you made a story specifically for `error`. Although imo as a user would like to see the implementation here. So:

```
  const fetchData = async () => {
    setFetchStatus('loading')
     try {
       const { data } = await axios.get('https://jsonplaceholder.typicode.com/users/1')
       window.setTimeout(() => {
         setFetchStatus('none')
         setLastUpdated(new Date(Date.now()))
         setFetchedData(data)
       }, 1000)
      } catch(error){
       setFetchStatus('error')
      }
  }
```

stories/refetch-data.stories.js
tlowande: I tried to replicate but I couldn't sam, when I do `new Date()` for the story, I don't see the component error. Same for the component itself, where I also cahanged to `new Date()` and didn't experience errors:
```  
const displayLastUpdated = () => {
    if (lastUpdated) {
      const startTime = lastUpdated
      const currentTime = new Date()
      ...
}
```

src/refetch-data.js
```diff
@@ -0,0 +1,184 @@
...
+    buttonRef.current.blur()
+  }
+
+  const handleKey = (event) => {
+    if (event.keyCode === 13 || event.keyCode === 32) {
+      fetchDataFunction()
+    } else {
+      buttonRef.current.blur()
+    }
+  }
```
tlowande, line 96: how do I trigger this? I tried pressing `return` and `space` but doesn't trigger. Was it made thinking of mobile support?

src/refetch-data.js
tlowande: seems to be working, just say my `14min ago` here.

stories/refetch-data.stories.js
SamSverko: Ohh, that was super helpful @tlowande thank you for the clarification :)

stories/refetch-data.stories.js
SamSverko: Using `const currentTime = new Date().toISOString()` for line 111 gives me the error. Thoughts @woozyking ?

src/refetch-data.js
```diff
@@ -0,0 +1,184 @@
...
+  const handleOpen = () => {
+    if (lastUpdatedDisplay) {
+      setOpen(true)
+    }
+  }
+
+  const displayLastUpdated = () => {
+    if (lastUpdated) {
+      const startTime = lastUpdated
+      const currentTime = new Date(Date.now())
```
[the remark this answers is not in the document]
SamSverko: This is missing the `.toISOString()` method. When that's added, it breaks with a very unhelpful error message.

src/refetch-data.js
SamSverko: This is for keyboard support, when you tab and focus onto the element.

tlowande (approved):

woozyking (approved):

PR #125, opened 2020-09-21
[G2M]styledCardContainer - background image adjustable to all ratios

tlowande:
_Problem_: in situations where `StyleCardComponent` is used with a variable ratio, the background image was not covering the whole div but actually repeating itself. Transition doesn't work with `background-size = 'cover'`
image

_solution_: the property `background-size` conditionally checks for the ratio of the component mounted and fits the image according to the biggest size. 
image

ps.: `scale(1.1)` method would zoom the whole card, including text. That is not the intended behavior

### DIFF
--- a/src/styled-card-container.js
+++ b/src/styled-card-container.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef, useState, useLayoutEffect } from 'react'
 import PropTypes from 'prop-types'
 import clsx from 'clsx'
 import { makeStyles } from '@material-ui/core/styles'
@@ -59,12 +59,11 @@ const useStyles = makeStyles((theme) => {
       boxSizing: 'border-box',
       height: '100%',
       width: '100%',
-      backgroundImage: (props) =>
-        `linear-gradient(${fade(theme.palette.grey[600], 0)}, #000), url(${
-          props.image
+      backgroundImage: ({ pattern }) =>
+        `linear-gradient(${fade(theme.palette.grey[600], 0)}, #000), url(${pattern.image
         })`,
       backgroundPosition: 'center center',
-      backgroundSize: '100% auto',
+      backgroundSize: ({ ratio }) => ratio ? '100% auto' : 'auto 100%',
       transition: 'background-size 0.6s ease-out',
       color: 'white',
       boxShadow: theme.shadows[1],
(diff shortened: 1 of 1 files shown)


SamSverko (approved):
This implementation looks familiar ;)

It looks great!

DoParkEQ:
I noticed that the current ratio based mechanism needs a width/height **ratio of the background image** into the equation to work as desired. 

Screen Shot 2020-09-22 at 1 44 44 AM

As an example, let's say there are two images. Image 1 has a landscape ratio and image 2 has portrait ratio. When `backgroundSize` is set to `100% auto` It will have empty space on the top and bottom for image 1.

To solve this, it needs to compute width and height ratio of the image then compare it with the container ratio.

src/styled-card-container.js
```diff
@@ -57,37 +57,44 @@ const useStyles = makeStyles((theme) => {
...
       width: '100%',
-      backgroundImage: (props) =>
-        `linear-gradient(${fade(theme.palette.grey[600], 0)}, #000), url(${
-          props.image
+      backgroundImage: ({ pattern }) =>
+        `linear-gradient(${fade(theme.palette.grey[600], 0)}, #000), url(${pattern.image
         })`,
       backgroundPosition: 'center center',
-      backgroundSize: '100% auto',
+      backgroundSize: ({ ratio }) => ratio >= 1 ? ' 100% auto' : 'auto 100%',
```
DoParkEQ, line 67: the conditional statement should look at both containerRatio and imageRatio to determine how the image should fit in the frame. 

I explained further why it needs w/h image ratio in a separate comment. 

There are four cases:

1. containerRatio > 1 && imageRatio > 1 --> `auto 100%`
2. containerRatio > 1 && imageRatio < 1 --> `100% auto`
3. containerRatio < 1 && imageRatio > 1 -->  `auto 100%`
4. containerRatio < 1 && imageRatio < 1 --> `100% auto`

src/styled-card-container.js
```diff
@@ -57,37 +57,44 @@ const useStyles = makeStyles((theme) => {
...
+        backgroundSize: ({ ratio }) => ratio >= 1 ? ' 105% auto': 'auto 105% ',
       },
     },
   }
 })
 
 const StyledCardContainer = ({ pattern, onClick, selected, children, ...rest }) => {
+  const [ratio, setRatio] = useState(0)
+  const ref = useRef(0)
+  useLayoutEffect(() => {
```
DoParkEQ, line 84: it can compute image ratio by this.
```jsx
    const img = new Image();
    img.src = pattern.image;  
    const { width, height } = img;
    const imageRatio = width/height;
```

src/styled-card-container.js
```diff
@@ -57,37 +57,44 @@ const useStyles = makeStyles((theme) => {
...
+      backgroundSize: ({ ratio }) => ratio >= 1 ? ' 100% auto' : 'auto 100%',
       transition: 'background-size 0.6s ease-out',
       color: 'white',
       boxShadow: theme.shadows[1],
       borderRadius: '10px',
       '&:hover': {
         transition: 'all .6s',
         boxShadow: theme.shadows[3],
-        backgroundSize: '105% auto',
+        backgroundSize: ({ ratio }) => ratio >= 1 ? ' 105% auto': 'auto 105% ',
```
DoParkEQ, line 75: The type of prop probably needs to be changed as it needs a result of the comparison between container ratio and image ratio to conditionally render.

src/styled-card-container.js
tlowande: instead of having so many combinations (in the cases above, `containerRatio` basically doesn't matter as `imageRatio` is taking priority) why don't we compare `imgWidth` vs `containerWidth`, same for height and then check what's bigger? Seems like it works, check the newest push

DoParkEQ (approved):

PR #130, opened 2020-10-07
[G2M]list/list-item - adjust style for status alignment on small items

tlowande:
- container shouldn't have `xs` attribute, which is pushing the chip out of the limits of the box when `width` is too small
<p>


SamSverko (commented):
Looks good to me!

Can we have the CHANGELOG updated to reflect this? e.g.:
```
### Fixed
- `<ListItem>` - Fixed status alignment style for small items.
```

SamSverko (approved):
🚀

PR #131, opened 2020-10-07
[G2M]list/list-item 

tlowande:
- adjust style for elements to work together on small items > previous fix #130 messed up with the title and description due to the size of the elements. `flex: 0` seems to overall do the job 🙏


SamSverko (approved):
🚀

PR #148, opened 2021-01-04
[G2M]Tabs - add vertical tabs option + return controlled tab value & onChange handler

hyx131:


src/tab-panels.js
```diff
@@ -24,18 +36,18 @@ const TabPanels = ({
     ))
 
   return (
-    <>
-      <TabsComponent value={value} onChange={onTabChange}>
+    <div className={TabsProps.orientation === 'vertical' ? classes.vertical : null}>
```
woozyking, line 39: would this be effectively equivalent?

```suggestion
    <div className={classes[TabsProps.orientation]}>
```

woozyking (approved):

woozyking:
this should be categorized as one of the "complex" component that may deserve it's own repo such as `lumen-tabs`

src/tab-panels.js
hyx131: yes it would! 😃

woozyking (approved):

woozyking:
perhaps update the PR title to reflect key points of updates. if this gets any deeper, merge and start another one to progress further

PR #153, opened 2021-01-14
[G2M] tabs - add forwardRef to use with drawers

tlowande:
**context:** ml-ui usecase to append a drawer anchored on the child component of that tab

**quirks**: commit 08c5198 uses storybook syntax BUT that is misleading. Under the hood Storybook seems to be merging the `WithRef.args` with default args of the component and then passing it to the component again. This means that we'd have to actually declare `forwardRef` not to get lost in translation. This gives the wrong idea that the array or references needs to be passed inside `forwardRef` when actual implementation needs it inside `ref`. So I went _tru_ with 5a2ead7 story syntax

image

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,8 @@
+import withRef from './with-ref'
+import TabPanels from './tab-panels'
+
+const TabPanelsWithRef = withRef(TabPanels)
+
 // untouched material-ui
 export { default as FormControl } from '@material-ui/core/FormControl'
 export { default as FormControlLabel } from '@material-ui/core/FormControlLabel'
@@ -29,6 +34,8 @@ export { default as TabPanels } from './tab-panels'
 export { default as TextField } from './text-field'
 export { default as Typography } from './typography'
 export { default as WidgetStats } from './widget-stats'
+export { default as withRef } from './with-ref'
+export { TabPanelsWithRef }
 
 // deprecated
 export { default as DataTable } from './deprecated/data-table' // deprecated by Table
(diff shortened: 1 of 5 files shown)


woozyking (approved):

src/index.js
```diff
@@ -29,6 +35,8 @@ export { default as TabPanels } from './tab-panels'
 export { default as TextField } from './text-field'
 export { default as Typography } from './typography'
 export { default as WidgetStats } from './widget-stats'
+export { default as withRef } from './with-ref'
+export { TabPanelsWithRef }
```
tlowande, line 39: is there a simpler way to export this with `@babel/preset-react`?

stories/tab-panels.stories.js
```diff
@@ -1,13 +1,50 @@
-import React from 'react'
+import React, { createRef } from 'react'
 
 import BuildIcon from '@material-ui/icons/Build'
 import BookmarksIcon from '@material-ui/icons/Bookmarks'
+import Drawer from '@material-ui/core/Drawer'
 
-import { TabPanels } from '../src/index'
+import { TabPanels, TabPanelsWithRef } from '../src/index'
+// import { withRef } from '../src/index'
```
tlowande, line 8: I left the comment out code to indicate that `TabPanelsWithRef` is the same thing as `withRef(TabPanels)` as in line 118

src/index.js
woozyking: lol `export TabPanelsWithRef`

stories/tab-panels.stories.js
woozyking: makes sense

woozyking (approved):
looks good.

next step shall we extract it out to `lumen-tabs` or something? it's now _complex_ enough with the alt handling of forward ref (and not)

hyx131 (approved):

PR #154, opened 2021-01-18
[G2M] devops - add auto-release

tlowande:
based on last 3 commits, if encounters a semver of any type, make a release


woozyking (approved):
i think we should reposition react-labs truly as a "labs" project where breaking changes _can_ happen (like 1.12 actually breaks for the image-slider's CSS import change), and speed up on moving the mature ones into `lumen-<complex-component>` and `lumen-ui` (for base components)

PR #155, opened 2021-01-21
[G2M] Extracted

tlowande:
- remove components extracted to EQWorks/lumen-ui
 ```
<Alert />
<Button />
<Checkbox />
<Chip />
<List />
<Loader />
<Radio />
<Switch />
<TabPanels />
<TextField />
<Typography />
<withRef />
<TabPanelsWithRef />
```

- update imports to use lumen-ui
- update peerdep on README
- update actions to have `yarn publish --access public`


tlowande:
> Would this cause `module not found` issue when react-labs are updated on Snoke side?

Probably. This means that from v2.0.0 and on any of the components of the list above will have to be imported from lumen-ui instead of react labs in snoke

woozyking (approved):

tlowande:
🚀

DoParkEQ (approved):

